Issue #38, opened 2026-06-25
Stale Cloudflare link: /workers/onboarding 404s in doctor + ONBOARDING.md (subdomain-claim step)


drbill-orbit:
The `cf_subdomain` step (doctor `missing_steps`, and `ONBOARDING.md:105`) points users to:

```
https://dash.cloudflare.com/<account_id>/workers/onboarding
```

which now returns **404** — Cloudflare moved the page. The Workers subdomain API confirms the current location in its own error message:

> You do not have a workers.dev subdomain. Please go to `https://dash.cloudflare.com/?to=/:account/workers/workers-and-pages`. Opening the Workers & Pages landing page for the first time will create a workers.dev subdomain automatically.

**Fix:** emit `https://dash.cloudflare.com/?to=/:account/workers/workers-and-pages` (or `/<account_id>/workers-and-pages`) for the `cf_subdomain` click step in both `bin/tdoc-doctor` and `ONBOARDING.md`.

**Env:** Cloudflare dashboard, June 2026.

Issue #37, opened 2026-06-25
macOS: publish can't read wrangler OAuth token (hardcoded ~/.wrangler path; current wrangler uses ~/Library/Preferences)


drbill-orbit:
On macOS with wrangler 4.105.0, a fresh `wrangler login` succeeds but `/tdoc publish` aborts with:

```
[tdoc] could not read wrangler OAuth token. Re-run 'wrangler login'.
```

`bin/tdoc-publish:140` (and `bin/tdoc-doctor`) hardcode `$HOME/.wrangler/config/default.toml`, but current wrangler uses `env-paths` and stores the token at `~/Library/Preferences/.wrangler/config/default.toml` on macOS. Because `wrangler whoami` works, the doctor reports `logged_in: true` while the token file read silently fails — so the failure only surfaces at publish time.

**Workaround:**

```bash
mkdir -p ~/.wrangler/config
ln -s ~/Library/Preferences/.wrangler/config/default.toml ~/.wrangler/config/default.toml
```

**Fix:** resolve the wrangler config path per-platform (macOS `~/Library/Preferences/.wrangler/...`, Linux XDG `~/.config/.wrangler/...`, legacy `~/.wrangler/...`) instead of assuming the legacy path.

**Env:** macOS, wrangler 4.105.0, node v26.

Issue #36, opened 2026-06-25
marketplace.json missing required `owner` field — `/plugin marketplace add` fails schema validation


drbill-orbit:
`/plugin marketplace add serenakeyitan/tdoc` fails immediately with:

```
Failed to parse marketplace file at .../.claude-plugin/marketplace.json:
Invalid schema: ... owner: Invalid input: expected object, received undefined
```

`.claude-plugin/marketplace.json` has no `owner` object, which the current Claude Code marketplace schema requires. This breaks the marketplace install path documented in the README and CONTRIBUTING.md, forcing users onto the `git clone` fallback.

**Fix:** add an `owner` object, e.g.

```json
"owner": { "name": "Serena Keyitan", "url": "https://github.com/serenakeyitan" }
```

**Env:** Claude Code (current), `marketplace.json` v0.1.44.

PR #42, opened 2026-06-27
fix(plugin): make manifest `repository` a string so Claude Code accepts it

drbill-orbit:
Thanks for the lightning-fast turnaround on #36, #37, and #38 — all three closed within a couple of days. Here's one more small one in the same spirit.

### The problem
When tdoc is installed as a Claude Code plugin/skill, Claude Code throws a manifest validation error on **every startup**:

```
Plugin tdoc has an invalid manifest file at ~/.claude/skills/tdoc/.claude-plugin/plugin.json.
Validation errors: repository: Invalid input: expected string, received object
```

Claude Code's plugin schema expects `repository` to be a **string** URL, but the manifest currently uses the npm-style object form:

```json
"repository": {
  "type": "git",
  "url": "https://github.com/serenakeyitan/tdoc"
}
```

The error flashes by quickly at launch, so it's easy to miss — but it fails validation each time.

### The fix
Flatten `repository` to the string URL form the schema accepts (same destination):

```json
"repository": "https://github.com/serenakeyitan/tdoc"
```

One line, `plugin.json` still valid JSON. Verified locally that this clears the startup error.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -27,8 +27,5 @@
     "prompt-native",
     "interactive"
   ],
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/serenakeyitan/tdoc"
-  }
+  "repository": "https://github.com/serenakeyitan/tdoc"
 }


serenakeyitan:
Merged — thanks @drbill-orbit! Verified the fix against the [official plugins reference](https://code.claude.com/docs/en/plugins-reference): the metadata table types `repository` as a **string** ("Source code URL"), and the npm-style object form we had was triggering the startup validation error you saw. Clean one-line fix, URL preserved.

That's your fourth solid contribution (#36/#37/#38 + this) — much appreciated. 🙏

PR #79, opened 2026-07-16
fix(overlay): table first column clipped inside overflow wrappers

drbill-orbit:
## Problem

The default doc template gives every table `margin-left: -14px` (`server/overlay.js`) to pull the first cell's text under the prose column. A negative left margin pushes the table's left edge **outside** its containing block — so any table inside an `overflow-x: auto` wrapper has its **first column clipped and unreachable by scroll** (the overhang sits before the scroll origin).

This is easy to hit because it's triggered by the project's *own* guidance: `SKILL.md` tells authors to wrap tables in `overflow-x: auto`, and the overlay's own `@media (max-width: 760px)` rule wraps them too. It's most visible when the first column is short (single-character cells), where the whole first column effectively disappears.

## Fix

Remove the negative left margin — tables sit flush in their container, with first-cell text indented by the cell padding, exactly like GitHub/Claude markdown tables.

```diff
- :where(body table) { border-collapse: separate; border-spacing: 3px; margin: 0 0 18px -14px; font-size: 16px; }
+ :where(body table) { border-collapse: separate; border-spacing: 3px; margin: 0 0 18px; font-size: 16px; }
```

**Trade-off:** first-cell text no longer hangs left under the prose — it's padded like a normal table cell. The "hang" can't be safely kept, because it's produced by the exact negative margin that clips. (If you'd rather preserve that alignment, zeroing the first column's left padding is a clip-safe alternative — happy to switch.)

## Test

Adds a gated Playwright regression test (`test/table-clip.test.js`) plus an isolated `wide-table` fixture (a table wrapped in `overflow-x: auto` with a single-character first column). It asserts the first cell never lands left of its wrapper.

Verified: **fails** on the old `-14px` margin (14px overhang), **passes** with the fix. Existing responsive/ui suites unchanged; offline suite green.

> Note: `table-clip.test.js` is registered in the `GATED` list. It pairs with the browser-CI PR below — once both land, `table-clip` should join that PR's `BROWSER` list so CI runs it too.


drbill-orbit:
Companion PR: #80 adds a `browser` CI job that runs the Playwright suites. Once both land, `table-clip.test.js` should be added to that PR's `BROWSER` list so CI runs it too.

serenakeyitan:
Reapplied onto current `main` as #128 — same CSS fix and regression test, rebased past the dark-mode overlay and later GATED list so it merges cleanly.

This original branch is still conflicting; please use #128.

serenakeyitan:
Superseded by #128 (reapplied onto current main).

PR #99, opened 2026-08-12
Require explicit GitHub OAuth client for published auth

yayashuxue:
## Summary
- remove the deprecated/personal GitHub OAuth client ID as a Worker/Vercel fallback
- disable published-doc auth when GITHUB_CLIENT_ID is unset instead of sending users to the old OAuth app
- update wrangler template and tdoc skill docs to require an org-owned OAuth app client ID
- add an offline regression test so the old client ID cannot be reintroduced into runtime config

## Validation
- node test/auth-config.test.js
- node test/vercel-shim.test.js
- node test/security.test.js
- git diff --check
- node test/run.js: 17/18 suites green; existing coverage.test.js bundle syntax check fails under local Node 18 because worker bundle contains export class CommentsStore parsed as CommonJS

## Rollout note
This PR prevents future deploys from silently using the old Serena/keyitan OAuth app. The live GitHub authorization screen will show tornado-doc only after an org-owned GitHub OAuth app is created/configured and its client ID is set as GITHUB_CLIENT_ID in the deployment environment.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -420,8 +420,10 @@ when `server/overlay.js` or `worker/worker.js` is newer than the bundled file
 and redeploys the Worker automatically so users get the latest overlay code.
 Set `TDOC_SKIP_WORKER_DEPLOY=1` to skip the redeploy (useful for batch uploads).
 
-On published docs, viewers sign in with GitHub (Device Flow, shared OAuth App
-`Ov23liZ1UAGOchvKPmlS`, scope `read:user`) before commenting.
+On published docs, viewers sign in with GitHub (Device Flow, scope
+`read:user`) before commenting. The deployment must configure an org-owned
+`GITHUB_CLIENT_ID`; if it is unset, published-doc sign-in is disabled rather
+than falling back to a personal OAuth app.
 
 Requires `jq`, plus `wrangler` (`npm i -g wrangler`) for the Cloudflare
 target or `vercel` (`npm i -g vercel`) for the Vercel target.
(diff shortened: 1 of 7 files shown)


serenakeyitan (commented):
## Product call

Agreed on direction. **Hold merge** until the org OAuth app exists and this is rebased.

Removing the personal `Ov23liZ1UAGOchvKPmlS` fallback is correct. GitHub’s consent screen shows the app owner; new deploys should not silently keep using Serena’s app. Fail-closed when `GITHUB_CLIENT_ID` is empty is the right default.

## Before merge

1. Create a **tornado-doc** org-owned GitHub OAuth app (Device Flow, `read:user`).
2. Set that client ID as `GITHUB_CLIENT_ID` on **tdoc.dev** (and document it for self-host).
3. **Rebase onto main.** `#110` already changed `injectOverlay` (nonce / owner-manage). This branch is 11 commits behind and will not apply cleanly.

Self-hosters who never set the var and relied on the template default will lose comment sign-in on the next redeploy. Please add one migration sentence in SKILL/README: set `GITHUB_CLIENT_ID` or published comments stay disabled.

Not approving yet only because of the rebase + live env. Code change itself LGTM.

yayashuxue:
Closing as outdated / superseded.

This branch is far behind main and conflicts. The org OAuth direction landed in #145 (`fix(auth): org OAuth SoT + landing redirects`): tornado-doc client via `shared/github-oauth.js`, no Serena personal fallback, template + Vercel consume the SoT.

No need to rebase this PR. Reopen only if something from this diff is still missing on main after #145.

serenakeyitan:
Checked `main` after #145 as requested. **No reopen.**

What landed covers this PR’s intent (and more):

- Org client `Ov23li1jJiBkS23x4O07` is the SoT in `shared/github-oauth.js`; the wrangler template and Vercel shim consume it.
- Personal `Ov23liZ1UAGOchvKPmlS` is gone from runtime config; `test/no-drift.test.js` blocks it coming back in SKILL/Vercel/template.
- `tdoc-publish` backfills existing BYOK `wrangler.toml` that still has the old id.
- SKILL.md documents the org app, not a personal fallback.

This branch’s fail-closed-when-empty approach is intentionally superseded by shipping the org SoT as the default. Nothing from this diff is missing on main.

PR #106, opened 2026-08-13
Expose runtime provenance and hash-based redeploy

yayashuxue:
## Summary
- add bundled runtime provenance to published docs via `window.__TDOC__.runtime`
- add `/api/runtime` for deployed Worker self-identification
- replace mtime-based Worker/Vercel redeploy detection with stored content hashes in `~/.tdoc/published.json`
- document the runtime provenance surface and hash redeploy behavior

## Verification
- `node test/runtime-provenance.test.js`
- `node test/coverage.test.js`
- `node test/run.js` (PASS: all 22 offline suites green)

## Notes
- This PR intentionally does not touch GitHub OAuth / #99.
- Do not merge until Julie explicitly approves; opened while she is sleeping for review only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,8 @@ Same document snapshots; different roles (see [AGENTS.md](AGENTS.md)):
 
 Overlay code is bundled into the hosted runtime at publish time from the skill
 checkout; local overlay edits only affect Local Studio until the next publish
-redeploy.
+redeploy. Published pages also expose the bundled runtime provenance at
+`/api/runtime` and in `window.__TDOC__.runtime`.
 
 ## How comments work
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -417,10 +417,12 @@ GitHub sign-in) work identically on either host. Caveats: no per-doc write
 serialization (Cloudflare uses a Durable Object for that) and a ~4.5 MB upload
 cap per doc (Vercel request limit).
 
-Subsequent runs upload the latest version of `<slug>`. The script also detects
-when `server/overlay.js` or `worker/worker.js` is newer than the bundled file
-and redeploys the Worker automatically so users get the latest overlay code.
+Subsequent runs upload the latest version of `<slug>`. The script compares a
+content hash of the Worker/overlay bundle against the last deployed hash in
+`~/.tdoc/published.json` and redeploys automatically when runtime code changed.
 Set `TDOC_SKIP_WORKER_DEPLOY=1` to skip the redeploy (useful for batch uploads).
+Published pages expose runtime provenance at `/api/runtime` and in
+`window.__TDOC__.runtime`.
 
 On published docs, viewers sign in with GitHub (Device Flow, shared OAuth App
 `Ov23liZ1UAGOchvKPmlS`, scope `read:user`) before commenting.
--- a/bin/tdoc-publish
+++ b/bin/tdoc-publish
@@ -220,11 +220,30 @@ bundle_worker() {
   OUT_DIR="$out_dir" node -e '
     const fs = require("fs");
     const path = require("path");
+    const crypto = require("crypto");
+    const cp = require("child_process");
     const skill = process.env.SKILL_DIR;
     const worker = fs.readFileSync(path.join(skill, "worker/worker.js"), "utf8");
     const overlay = fs.readFileSync(path.join(skill, "server/overlay.js"), "utf8");
-    // Replace the whole declaration so JSON.stringify safely escapes everything.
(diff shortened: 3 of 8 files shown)


serenakeyitan:
Post-merge review: this was merged before the requested review arrived. The full offline suite is green (`node test/run.js`: all 22 suites), but I found four release-state bugs that remain on current `main` (`9f0eff4`).

1. **[P1] The new config updates expose the upload token through a default-permission temp file.** [`bin/tdoc-publish:637-638`](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L637-L638) and [697-698](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L697-L698) redirect the entire secret-bearing `published.json` to the predictable `published.json.tmp` before `chmod 600`. With the normal `022` umask, that file is initially `0644`; an interrupted/failed rename can leave the token readable by other local users. Use one atomic config-update helper that creates a unique mode-0600 temp file before writing.

2. **[P1] Vercel can persist a source hash for a different staged deployment.** [`vercel_bundle_sha`](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L281-L305) hashes the current source tree, while [`sync_vercel_app`](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L466-L475) only copies into a persistent directory and never removes deleted/renamed files. A removed API route or library remains in the deployment, then the new source hash is stored and future publishes skip the corrective deploy. Build from a clean managed staging tree and hash the exact staged artifact.

3. **[P1] First-time setup always deploys twice, and the second Vercel deployment is not smoke-tested.** Both setup functions deploy, then write configs without the new hash fields ([Cloudflare](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L418-L440), [Vercel](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L546-L579)). Execution falls through, sees an empty stored hash, and immediately redeploys. On Vercel, the first deployment is checked via `/api/ping`; the second is recorded without a health check and can leave config pointing at a broken release. Persist the initial hash or skip the subsequent redeploy block for that invocation, with a hermetic “exactly one deploy” test for each platform.

4. **[P1] The redeploy decision trusts local state instead of the deployed runtime.** [`DEPLOYED_BUNDLE_SHA` comes only from `published.json`](https://github.com/tornado-doc/tdoc/blob/169ba3edf5bb812b2e8f6960bd815652d10d7db5/bin/tdoc-publish#L621-L625). An out-of-band rollback/dashboard deployment or restored stale config can therefore make publish report success while remote code remains stale. The new `/api/runtime` endpoint provides the remote evidence needed here; treat the local hash as a cache and verify remote provenance before skipping deployment. For Vercel, route `/api/runtime` before storage initialization so it still works when Blob/Redis is broken.

Two smaller follow-ups: hash only the files `sync_vercel_app` actually deploys (the current recursive walk includes `vercel/README.md` and ignored/generated directories), and remove or consume/gitignore the unused `_worker.bundled.sha256` sidecar. The new tests currently assert source strings rather than executing these first-run, staging, permission, and failure paths, which is why all tests pass despite the issues above.

---

serenakeyitan (commented):
Formal review to close the outstanding request. I independently re-read the PR diff and re-checked current `main` (`9f0eff4`).

The change itself is the right direction: content-hash redeploy beats mtimes, `/api/runtime` + `window.__TDOC__.runtime` is a useful published-page contract, and `/api/ping` stays a minimal health check. Offline suites I ran are green (`test/runtime-provenance.test.js`, `test/coverage.test.js`).

I am not approving post-merge because the four release-state bugs in https://github.com/tornado-doc/tdoc/pull/106#issuecomment-5292207834 are still present on `main`. Please treat that comment as the review record and land a follow-up. One extra implementation note on finding 4: even after publish consults `/api/runtime`, the hashes are not comparable as written. `runtime.bundle_sha` is a 12-char prefix of `sha256(worker + "\0" + overlay)`; `vercel_bundle_sha256` hashes the entire `vercel/` tree, and the stored Cloudflare hash is the full digest. The skip decision needs one shared digest.

---

PR #107, opened 2026-08-13
Fix access policy write validation

yayashuxue:
## Summary
- reject invalid access enum values and malformed allowlists on write paths instead of silently normalizing them
- keep legacy/corrupt read fallback in normalizeAccess, but require strict values for PATCH /api/doc/access and upload meta.access
- validate upload access before writing R2 document bytes, so malformed policy fails closed without partial doc writes

## Verification
- node test/access.test.js
- node test/remote-access-route.test.js
- node test/run.js
- mutation check: removing visibility validation makes access.test.js fail

### DIFF
--- a/worker/worker.js
+++ b/worker/worker.js
@@ -158,6 +158,51 @@ function accessFromMeta(meta) {
   return normalizeAccess(has ? meta.access : null, { legacy: !has });
 }
 
+function validateAccessWrite(access) {
+  if (!access || typeof access !== 'object' || Array.isArray(access)) {
+    return { error: 'access object required' };
+  }
+  const keys = Object.keys(access);
+  const unknown = keys.filter((k) => !ACCESS_PATCH_FIELDS.has(k));
+  if (unknown.length) return { error: 'invalid_access_field', fields: unknown };
+
+  const out = {};
+  if ('visibility' in access) {
+    if (!ACCESS_VISIBILITIES.has(access.visibility)) {
+      return { error: 'invalid_access_value', field: 'visibility' };
+    }
+    out.visibility = access.visibility;
+  }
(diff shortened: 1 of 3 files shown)


serenakeyitan (commented):
Post-merge review found one P1 access-policy regression. The existing 22-suite run is green, but the new upload test checks source ordering rather than executing this branch; the failing behavior is also still present on current main. Agent note: this reply was prepared and posted by breeze running locally for the active account.

---

worker/worker.js
```diff
@@ -2277,6 +2322,27 @@ export default {
...
+        incoming = (meta && typeof meta === 'object') ? { ...meta } : {};
+        let prev = null;
+        try {
+          const prevRaw = await env.META.get(`meta:${slug}`);
+          if (prevRaw) prev = JSON.parse(prevRaw);
+        } catch {}
+        if (!incoming.access && prev && prev.access) {
+          incoming.access = prev.access;
+        }
+        if (incoming.access) {
```
serenakeyitan, line 2338: [P1] JSON-present falsy access values bypass strict validation. Both preservation and validation use truthiness here. For a new slug, `meta.access` set to `null`, `false`, `0`, or `""` skips `validateAccessWrite`, returns HTTP 200, writes R2, persists malformed metadata, and `accessFromMeta` then applies the legacy public/public fallback. This violates the PR’s fail-closed write contract. Please distinguish property absence from a present falsy value: inherit the previous policy only when `access` is absent, and validate whenever it is present. Add a route-level `fetch` test with `{ meta: { access: null } }` that asserts HTTP 400 plus zero `DOCS.put`/`META.put` calls. Agent note: this reply was prepared and posted by breeze running locally for the active account.

---

PR #108, opened 2026-08-13
Keep docs centered when comment pins are visible

yayashuxue:
## Summary

Fixes the published reader layout regression Julie called out on `tdoc-hosted-product-model`: wide comment mode was reserving a 360px right rail on `body`, so the document stayed centered in the reduced reading area but appeared 180px left of the viewport center at 1440px.

This PR treats comment pins/cards as provider chrome instead of document layout:

- keep the document centered in the viewport when wide pinned comments are visible
- let the existing layout evaluator switch to the drawer when comment UI cannot fit beside the article
- make narrow-mode top-bar controls follow `body.tdoc-narrow`, not only viewport media queries
- record the provider-enforcement rule in `AGENTS.md`
- update browser tests to assert viewport centering and to open hidden cards through visible pins
- harden `publish.test.js` to use an isolated `127.0.0.1` port instead of fixed `localhost`

## Root Cause

PR #97 centered the article in the reader area by applying `padding-right: 360px` to `body.tdoc-has-comments`. That was correct for the older always-visible comment rail, but after wide mode moved to pins with cards hidden until opened, the same padding still shifted every commented doc left even though there was no persistent rail to reserve.

## Validation

- `node test/responsive.test.js`
- `node test/dimensions-audit.js`
- `node test/publish.test.js`
- `node test/ui.test.js`
- `node test/run.js`
- `node test/run.js --all`

`node test/run.js --all` passes all 26 suites locally.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,1 +1,2 @@
 Remote storage is source of truth. Local HTML is disposable. Local skill is authoring/scaffold.
+Published reader invariants are provider-enforced in overlay/worker code and tests, not left only to author HTML or prompts.
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -81,11 +81,11 @@
   body { padding-top: 44px !important; padding-bottom: 24px; -webkit-user-select: text; user-select: text; }
   body .tdoc-bar, body .tdoc-bar *, body #tdoc-comment-layer, body #tdoc-comment-layer *, body #tdoc-pin-layer, body #tdoc-pin-layer *, body .tdoc-cluster-pop, body .tdoc-cluster-pop *, body .tdoc-hover-outline, body .tdoc-comment-pill, body .tdoc-emoji-picker, body .tdoc-secondary-menu, body .tdoc-anchor-mark.tdoc-anchor-mark-element, body .tdoc-drag-marquee, body .tdoc-modal, body .tdoc-modal * { -webkit-user-select: none !important; user-select: none !important; }
   body .tdoc-modal .code, body .tdoc-modal textarea, body .tdoc-modal input { -webkit-user-select: text !important; user-select: text !important; }
-  /* Reserve the comment rail on the right. The body content box becomes the
-     reader area, so article margin:auto centers content in that area exactly
-     instead of using a compensating left gutter. */
-  body.tdoc-has-comments:not(.tdoc-narrow) { padding-right: 360px !important; padding-left: 0 !important; }
-  body.tdoc-narrow { padding-right: 0 !important; }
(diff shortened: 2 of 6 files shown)


serenakeyitan:
Post-merge review: changes requested. This PR was already merged before the queued review ran, so these need a fix-forward follow-up.

1. **[P1] Open cards overflow and overlap their pins at common desktop widths.** `gutterGeometry()` uses `cardWidth = 280`, but `.tdoc-margin-comment` is content-box sized, so its actual outer width is 306px. At 1366px I reproduced: article `323..1043`, pin `1055..1083`, card `1074..1380`. The card overlaps the pin and extends 14px past the viewport. The 300px `MIN_COLUMN_WIDTH` threshold does not reserve the full pin + gaps + outer-card footprint. Please calculate the threshold/clamp from the real outer width (or make the card border-box) and cover an opened card in the dimension audit.

2. **[P1] Wrapperless documents stay in wide mode and let cards cover prose.** `getArticleMetrics()` returns real fallback prose geometry with `el: document.body`, but `evaluateLayout()` replaces that width/right edge with `Infinity`/`0`. At 1024px I reproduced `narrow=false`, prose `152..872`, and card `732..1038`: 140px of prose is covered and the card overflows. Preserve the fallback geometry when deciding narrow mode and add a wrapperless fixture.

3. **[P2] Wrapperless mobile docs lose their reading gutter.** The new `body.tdoc-narrow { padding-left: 0 !important; }` overrides the provider's own 24px wrapperless-body padding. At 375px I reproduced computed left/right padding of `0px` and prose spanning `0..375`, even though this behavior does not depend on comments. Scope the reset to wrapped docs or preserve the wrapperless gutter.

4. **[P2] The new regression guard is not actually a standard gate.** `dimensions-audit.js` is absent from `test/run.js`, so neither `npm test`, `node test/run.js --all`, nor CI runs the centering assertion. The assertion also explicitly accepts the obsolete `(innerWidth - 360) / 2` center when `tdoc-pins` is missing, masking a regression of the new invariant. Add the suite to the gated runner, require `tdoc-pins` for wide commented layouts, and always assert viewport centering.

5. **[P2] Authoring guidance is stale.** `SKILL.md:590` and `skills/tdoc/SKILL.md:588` still tell authors that comments anchor the article left for a right-side card column, directly contradicting the new provider-enforced centered layout. Update both copies and guard the wording in `agent-md.test.js`.

Verification: `node test/run.js --all` passed all 26 registered suites with Playwright available; direct runs of `dimensions-audit.js` (26 widths), `responsive.test.js`, `ui.test.js`, and `publish.test.js` also passed. The reproductions above show the current fixtures and audit do not cover the failing geometry.

---

PR #109, opened 2026-08-13
Add hosted publish target

yayashuxue:
## Summary

Adds the JUL-30 hosted/out-of-box publish slice: `/tdoc publish --platform hosted <slug>`.

This gives the publish CLI a tdoc-managed target that skips user-owned deploy setup:

- no Wrangler requirement
- no Vercel CLI requirement
- no R2/KV/Worker/Vercel project setup by the user
- no OAuth app setup by the user
- defaults hosted base to `https://tdoc.dev`, overridable with `TDOC_HOSTED_BASE`
- first hosted publish calls `POST /api/hosted/token` to get an account-scoped upload token
- persists `platform: "hosted"`, `base`, `public_host`, `account_id`, and `upload_token` in `~/.tdoc/published.json`
- reuses the existing `/api/upload` payload path, so access policy, versions, comments, and pure-publish behavior stay shared with self-host
- teaches `tdoc-pull` and `tdoc-unpublish` to use hosted `.base` instead of deriving a Cloudflare workers.dev URL

Server-side enforcement added in this PR:

- hosted tokens are stored as hashes in `META`, not as raw token keys
- provider admin token still works for existing single-tenant/self-host admin flows
- hosted tokens can create a new slug, then only mutate/delete/access-update/comment-wipe/agent-reply docs stamped with their `account_id`
- first hosted slug claim is serialized in the per-slug Durable Object (`hostedOwner`) rather than KV read-then-write
- hosted upload persists ownership even when the client omits `meta`
- R2 existence-check uncertainty fails closed before DO owner claim or R2 write
- same hosted owner can retry and repair ownership metadata after a partial R2-success / META-failure upload
- non-create hosted writes fail closed when remote ownership is absent; legacy/orphan docs with no meta cannot be deleted/wiped/replied to by arbitrary hosted tokens
- attempts to take an existing non-hosted slug return `slug_taken`; attempts to mutate another hosted account's slug return `not_doc_owner`
- provider can keep self-serve registration closed via env; `TDOC_HOSTED_REGISTRATION=1` enables `/api/hosted/token`

## Why

The existing `tdoc-publish` product story was still self-host first: first publish walked Cloudflare or Vercel resource setup. Julie's JUL-30 direction is hosted/out-of-box first, with self-host as an advanced path. This PR makes hosted publish self-provisioning without putting the provider-wide upload secret on user machines.

## Explicit Non-Goals

This does not add reader/comment identity beyond the existing GitHub flow. Hosted v1 remains link-share oriented; private allowlist and comment identity can be a later hosted-account layer.

This PR also does not change the default platform from Cloudflare yet. Hosted can become default after the provider intentionally enables registration and we smoke it on `tdoc.dev`.

## Validation

- `bash -n bin/tdoc-publish bin/tdoc-pull bin/tdoc-unpublish`
- `node test/cli.test.js`
- `node test/hosted-oob.test.js`
- `node test/hosted-oob-behavior.test.js`
- `node test/remote-access-route.test.js`
- `node test/publish.test.js`
- `node test/run.js` — all 24 offline suites green
- `node test/run.js --all` — all 28 suites green

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,10 @@ Publishing has one hosted target and two self-host targets; pick one on your
 first publish and it sticks (saved in `~/.tdoc/published.json`):
 
 - **Hosted** — `/tdoc publish --platform hosted <slug>` uploads to a
-  tdoc-managed host such as `tdoc.dev`. This path skips user deploy setup. In
-  hosted v1, the upload token is configured out-of-band with
-  `TDOC_HOSTED_UPLOAD_TOKEN`; account/token issuance is the next product layer.
+  tdoc-managed host such as `tdoc.dev`. This path skips user deploy setup. On
+  first use, the hosted service issues an account-scoped upload token and the
+  CLI stores it in `~/.tdoc/published.json`; that token can only mutate docs it
+  owns.
 - **Cloudflare (default)** — Worker + R2 + KV, with a Durable Object
   serializing concurrent comment writes. The most battle-tested target.
 - **Vercel** — `/tdoc publish --platform vercel <slug>` deploys a catch-all
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ Claude-specific tools are unavailable. See [Using tdoc with Codex](#using-tdoc-w
 ```
 You:  /tdoc new "an explainer with a slider showing how interest compounds"
 Claude: <generates doc, opens it locally>
-You:  /tdoc publish
-Claude: https://tdoc.yourname.workers.dev/d/compound-interest/v/1
+You:  /tdoc publish --platform hosted
+Claude: https://tdoc.dev/d/compound-interest/v/1
 ```
 
 Anyone with the link reads it instantly and comments on any sentence, image, or
@@ -79,7 +79,7 @@ What is *not* yet first-class on Codex: native slash-command registration (`/tdo
 |---|---|
 | `/tdoc new <prompt>` | Generate a new doc + open locally |
 | `/tdoc edit <slug>` | New version from open comments; replies on each with ✅/🟡/❓ status |
-| `/tdoc publish <slug>` | Deploy to your Cloudflare Worker (or Vercel, `--platform vercel`), get a public URL |
+| `/tdoc publish <slug>` | Publish a doc and get a public URL; hosted service is available with `--platform hosted`, self-host remains Cloudflare/Vercel |
 | `/tdoc pull <slug>` | Sync comments from the published doc back to local |
 | `/tdoc fork <slug>` | Copy a doc to a new slug |
 | `/tdoc unpublish <slug>` | Remove a published doc from your Worker |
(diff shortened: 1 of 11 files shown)


serenakeyitan (changes requested):
## Product call

**Default is hosted / `tdoc.dev`.** `/tdoc publish` with no flag should go there. Cloudflare and Vercel stay available as self-host:

- `--platform cloudflare`
- `--platform vercel`

The README example that shows `tdoc.dev` is the right happy path. Do not treat hosted as a side path.

## Request changes

1. **Rebase onto main (#110) and keep one combined write gate.**
   `#110` already shipped `authorizeOwnerMutation` (owner session **or** upload token) and the old `requireUploadAuth` contract (`Response | null`, truthy = 401). This PR changes that helper to `{ok, actor, response}` and does not have `authorizeOwnerMutation`.
   A naive rebase fails in one of two ways:
   - leftover `if (unauth)` rejects every valid token, or
   - leftover “token passed ⇒ global owner” lets **any hosted token** DELETE/PATCH someone else’s doc.
   The merged gate has to be: `TDOC_OWNER` session **or** `(requireUploadAuth + requireDocWriteAccess` scoped to that slug`)`. Do not keep only one side.

2. **Release `hostedOwner` on delete/unpublish.**
   `claim_owner` runs before access validation and R2/META writes. A 400 still parks the slug. `DELETE /api/doc` clears R2/meta/comments and never deletes the Durable Object owner key, so a deleted slug stays reserved forever. Add a `release_owner` op and call it from delete. Please add a delete-then-reclaim test (original owner can republish; a second token cannot squat).

3. **Do not flip `TDOC_HOSTED_REGISTRATION=1` on tdoc.dev in this PR.**
   `POST /api/hosted/token` is unauthenticated, unlimited, non-expiring, and sits behind `Access-Control-Allow-Origin: *`. Fine while the flag is off (tdoc.dev currently 404s this route). Not fine to turn on yet.
   The CLI/docs can default to hosted; the live service staying closed is an ops hold. Until registration is on, a no-flag publish should fail with a clear “hosted signup isn’t open, use `--platform cloudflare` or `--platform vercel`” — not a generic HTTP error.

## Follow-ups (not merge blockers if registration stays off)

- Access-policy `owner` is still `TDOC_OWNER`, not the hosted publisher. `--visibility private` locks the publisher out of their own doc.
- Upload with no `access` block still persists legacy **public**, not the advertised unlisted default.
- Same-origin user HTML + `tdoc_sid` is an existing launch-blocker; hosted makes it multi-tenant. Out of scope here, but it is why registration stays off.

CI is green and the fail-closed ownership tests on the happy path are real. Rebase + owner release + hosted-as-default docs, then this can land with registration still closed.

yayashuxue:
Superseded by https://github.com/tornado-doc/tdoc/pull/137.

#109 was based on an old main (pre-#110 session gate). Rebasing in place would either keep leftover `if (unauth)` 401s or treat any hosted token as a global owner. #137 is a fresh branch off current main with Serena’s blockers: combined write gate, `release_owner` on delete, hosted as the default CLI target, and tdoc.dev registration still closed.

PR #110, opened 2026-08-14
JUL-36/37: owner manage UX — no browser token (CSP), /me clean catalog, doc-page Share panel

yayashuxue:
## What & why

Removes the browser **admin-token** requirement for owner management and makes the owner surfaces clean. julie's asks: "access control UI 不好看" + "不能有 admin token".

### The core move (why it's safe to drop the token)
Published docs are arbitrary same-origin HTML — a doc `<script>` could ride the owner's session cookie to silently delete docs (confused-deputy). That's the *only* reason browser writes needed an explicit token. Fix: a per-response **CSP** (`script-src 'nonce-<n>' 'strict-dynamic'`) blocks all author scripts/inline-handlers while the overlay's own nonced scripts run → no untrusted code in the browser → cookie/session auth is safe → the browser token is removed. CLI publish keeps its bearer token (CLIs have no origin/scripts).
0 of 36 published doc versions use `<script>`, so zero breakage.

### Changes
- **CSP** on every doc-serving response — doc-view **and** `/export`/`/fork` (same author-HTML surface; covering only `/d/` would be bypassable via `/export`). Nonce is fresh-random per response; author HTML never gets it. Fixed one inline `onerror` in `avatarHTML`.
- **`authorizeOwnerMutation`** (session **OR** upload token) on `DELETE /api/doc` + `PATCH /api/doc/access` — browser uses the session (no token), CLI keeps the token; ownership re-checked inside each route.
- **`/me` → clean catalog**: dropped the visibility/history/commenting/allowed dropdowns + admin-token input; response no longer carries access data (no `allowed_users`). Delete tucked behind a quiet `⋯` overflow (airy border-bottom list, pre-#104 aesthetic).
- **Doc page → "Share settings" panel**: visibility / history / commenting / allowed-users, no token field, PATCH via same-origin cookie. Styled confirm modals kept; modal width capped so it never goes edge-to-edge (narrow-mode margin fixed).

### Verification (self-run)
- **24 offline suites green** + **csp-xss browser proof 7/7**: author `<script>`/`onclick` never execute (`window.__XSS__` undefined) with a **real CSP violation logged**; overlay + commenting fully work under the policy; **negative-mutation checked** (remove the CSP header → the XSS tests go red); **nonce hardening** — two requests get different nonces, and an author script carrying a *real captured* nonce is still blocked (per-response-random, non-replayable).
- Preview screenshots (clean `/me`, Share panel, subtle `⋯` delete) shared with julie for taste sign-off.

Reviewer (@小cc) will independently reproduce the CSP block.

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -112,6 +112,9 @@
     --td-accent-ring-soft: rgba(22,82,240,0.18);
     --td-accent-wash: rgba(22,82,240,0.06);
     --td-accent-tint: #e8eeff;
+    --td-danger: #b42318;
+    --td-danger-hover: #931c14;
+    --td-danger-tint: #fdeceb;
     --td-ground: #fff;
     --td-ink: #1a1a1a;
     --td-heading: #1a1a1a;
@@ -543,6 +546,23 @@
   .tdoc-modal .danger { color: #c33; font-size: 13px; }
   .tdoc-modal code { background: #f5f6f8; padding: 1px 5px; border-radius: 3px; }
 
+  /* Manage-doc modal (JUL-36: Delete / Unpublish / visibility switch). */
+  .tdoc-modal .manage-section { margin: 16px 0; }
+  .tdoc-modal .manage-section:first-of-type { margin-top: 4px; }
+  .tdoc-modal label.field { display: block; font-size: 12px; color: #666; margin: 0 0 4px; font-weight: 600; }
+  .tdoc-modal input[type="password"] { width: 100%; box-sizing: border-box; border: 1px solid #ccc; border-radius: 6px; padding: 8px 10px; font: inherit; }
+  .tdoc-modal input[type="password"]:focus { outline: none; border-color: var(--td-accent); }
+  .tdoc-seg { display: inline-flex; border: 1px solid #ddd; border-radius: 8px; overflow: hidden; }
+  .tdoc-seg button { border: none; border-radius: 0; padding: 7px 14px; background: #fff; color: #444; }
+  .tdoc-seg button + button { border-left: 1px solid #ddd; }
+  .tdoc-seg button.active { background: var(--td-accent); color: #fff; }
+  .tdoc-modal .manage-hint { font-size: 12px; color: #888; margin: 6px 0 0; }
+  .tdoc-modal .manage-action { width: 100%; text-align: left; }
+  .tdoc-modal .manage-action.danger-btn { color: var(--td-danger); border-color: #f1b8b2; }
+  .tdoc-modal .manage-action.danger-btn:hover { background: var(--td-danger); color: #fff; border-color: var(--td-danger); }
+  .tdoc-modal button.danger { background: var(--td-danger); border-color: var(--td-danger); color: #fff; }
+  .tdoc-modal button.danger:hover { background: var(--td-danger-hover); border-color: var(--td-danger-hover); }
+
   /* Bar collapse breakpoints — tied to viewport width, not layout class.
      The bar progressively hides elements as the viewport tightens, so it
      stays elegant at every size.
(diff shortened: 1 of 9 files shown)


serenakeyitan:
Post-merge review found security and correctness blockers that are still present on `main`. I recommend an immediate follow-up (or rollback of session-authorized mutations) before treating the browser owner-management flow as safe.

1. **[P0] Privileged owner controls share the author-controlled DOM/CSS trust boundary.** Raw author HTML is emitted before the overlay, author CSS remains unrestricted, and privileged elements/handlers are resolved with global `document.getElementById` calls (`server/overlay.js:939,2437-2442,2526,2545-2617`). An author can predeclare the same IDs and use hostile CSS to capture or overlay management controls. A headless-Chrome proof using this branch's CSP reached a session-authorized `DELETE /api/doc?slug=victim-doc` through colliding author elements. Move provider UI outside the author document (prefer a trusted parent with author content in a sandboxed, cookie-less iframe/origin); at minimum isolate provider UI in a closed shadow root, retain direct references, and add hostile-ID/CSS browser tests.

2. **[P0] The owner UI is frameable, enabling session-cookie clickjacking.** `cspHeader()` has no `frame-ancestors`, and `/me` returns HTML with neither CSP nor `X-Frame-Options` (`worker/worker.js:875-876,1956-1964`). A browser proof confirmed that an author page could frame `/me` and that the request carried `tdoc_sid`; transparent UI redressing can then drive the owner through the overflow, Delete, and confirmation clicks. Add `frame-ancestors 'none'` plus `X-Frame-Options: DENY` to privileged/provider HTML, and add a browser regression test. See [MDN's clickjacking guidance](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/Clickjacking).

3. **[P1] Blocking all author JavaScript breaks tdoc's documented content contract instead of establishing a safe boundary.** The authoring skill explicitly requires inline JS and interactive simulations (`SKILL.md:224-231`), and its canonical template includes an inline `<script>` (`SKILL.md:558-573`). The new nonce-only CSP intentionally blocks every author script (`worker/worker.js:865-876`), including local/forward-generated interactive docs. The fact that 0/36 currently stored versions use scripts does not reconcile that contract. Keep token-only mutations until author content and provider controls/cookies are isolated; do not rely on disabling the product's documented interactivity as the authorization boundary.

4. **[P1] Access mutations can report success after failure and race destructive operations.** `patchAccess()` catches and swallows `ownerFetch` failures, while Private/Unpublish callers continue with "Done", close, and render Private (`server/overlay.js:2531-2565,2597-2605`). Separately, PATCH loads metadata and later unconditionally writes it, while DELETE independently removes the same key; a delayed PATCH can recreate stale metadata after deletion (`worker/worker.js:2541-2547,2555-2573`). Return/rethrow failures and update UI only from confirmed server state. Serialize every per-slug administrative mutation through one Durable Object or use a tombstone/version precondition. This also avoids relying on Workers KV for concurrent writes, which [Cloudflare documents as overwrite-prone and limited to one write per second per key](https://developers.cloudflare.com/kv/api/write-key-value-pairs/).

5. **[P1] "Unpublish" has two incompatible, data-loss-prone meanings.** The new panel's Unpublish only PATCHes visibility to private and promises that versions/comments are kept (`server/overlay.js:2597-2605`), while the existing `tdoc-unpublish` CLI sends DELETE and permanently removes all versions/comments (`bin/tdoc-unpublish:39-40`). Rename the reversible UI action to "Make private", or change the CLI/legacy UI contract to PATCH private and reserve "Delete" for hard deletion.

6. **[P1] The only browser proof for the new security boundary is not required by CI.** `test/csp-xss.test.js` is in the gated suite, while the workflow runs only the dependency-free offline suite. The local browser test also skipped because Playwright was unavailable. Add a required browser job that exercises the Worker response path, CSP enforcement, framing protection, hostile DOM/CSS collisions, and session-authorized mutations.

Verification: `npm test` passed all 24 offline suites. That does not cover the browser-only security boundary above.

---

PR #113, opened 2026-08-15
Fold Share settings into the single Share button

yayashuxue:
## What & why

#110 added a second owner entry (Share settings in the profile menu) next to the existing Share button. One Share control is enough: owners get copy-link plus access settings in the same panel; everyone else still gets copy-link only.

## Checklist

- [x] `npm test` passes locally (offline suite).
- [ ] If I touched anything browser-facing (`server/overlay.js`, `server/server.js`)
      or the worker (`worker/worker.js`), I ran `npm run test:all` with Playwright
      installed (`npm i -D playwright && npx playwright install chromium`).
      (Share UI is published-only; offline suite covers the overlay source guards. Playwright Share tests updated but not re-run against a live worker.)
- [x] If I edited `SKILL.md`, I copied it to `skills/tdoc/SKILL.md` so the two
      stay in sync (`cp SKILL.md skills/tdoc/SKILL.md`). They must match — the
      plugin-mode install reads the copy. (n/a — SKILL.md unchanged)
- [x] If I changed the version, I bumped `VERSION` **and** `.claude-plugin/plugin.json`
      together (the manifest test enforces this). (n/a)

## Test plan

- [ ] On a published owner view, click Share: copy link at top, visibility/history/commenting/unpublish/delete below. Profile menu has no Share settings.
- [ ] Same URL signed out / incognito: Share is copy-link only.
- [ ] Confirm the stale CLI `/tdoc unpublish` hint is gone from the reader Share modal.

## Security note

No new mutation path. Owner settings still render only when `cfg.ownerManage` is present (server-gated per request); `showManageModal()` still bails before creating DOM without it. Share remains visible to all published viewers because copy-link is not owner-only.

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -561,8 +561,8 @@
   .tdoc-modal .danger { color: #c33; font-size: 13px; }
   .tdoc-modal code { background: #f5f6f8; padding: 1px 5px; border-radius: 3px; }
 
-  /* Share panel (JUL-36, reworked 2026-08-13: visibility/history/commenting/
-     allowed_users + Delete/Unpublish — no admin token, session-authorized). */
+  /* Share panel (copy link for everyone; owners also get visibility/history/
+     commenting/allowed_users + Delete/Unpublish — session-authorized). */
   .tdoc-modal .manage-section { margin: 16px 0; }
   .tdoc-modal .manage-section:first-of-type { margin-top: 4px; }
   .tdoc-modal label.field { display: block; font-size: 12px; color: #666; margin: 0 0 4px; font-weight: 600; }
@@ -720,7 +720,7 @@
     </div>`;
 
   const primaryCtaHtml = isFork ? '' : (isPublished
-    ? `<button id="tdoc-share-btn" class="primary" title="Share link" aria-label="Share">
+    ? `<button id="tdoc-share-btn" class="primary" title="Share" aria-label="Share">
          <svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M10 13a5 5 0 0 0 7.54.54l3-3a5 5 0 0 0-7.07-7.07l-1.72 1.71"/><path d="M14 11a5 5 0 0 0-7.54-.54l-3 3a5 5 0 0 0 7.07 7.07l1.71-1.71"/></svg>
          <span>Share</span>
        </button>`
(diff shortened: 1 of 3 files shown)


server/overlay.js
```diff
@@ -2365,42 +2357,39 @@
     };
   }
   function showShareModal() {
+    // One Share button: owners get copy-link + access settings in the same
+    // panel; everyone else gets copy-link only. No separate "Share settings".
+    if (cfg.ownerManage) { showManageModal(); return; }
```
serenakeyitan, line 2362: Follow-up: this is now the only owner entry, and it keys only off boot-time `cfg.ownerManage`. Sign out already sets `isOwner = false` and tears down the identity chip, but it leaves `cfg.ownerManage` in place — so Share after logout still opens Unpublish / Delete / the allowlist. Mutations 401 (session cookie is gone); the leak is the UI plus `allowed_users` / counts staying on screen.

The previous menu item was gated on `isOwner && cfg.ownerManage` and vanished with the chip. Either clear `cfg.ownerManage` in the sign-out handler, or dispatch on `isOwner && cfg.ownerManage` so the live flag the logout path already updates is what the button consults.

serenakeyitan (approved):
The fold is the right product change: one Share control, owners get copy-link plus access settings, everyone else still gets copy-link only. Server gating of `ownerManage` is unchanged, `showManageModal()` still bails before creating DOM, and dropping the reader-facing `/tdoc unpublish` hint is correct — that path never worked in the browser. Reusing `#tdoc-share-url` / `#tdoc-share-copy` / `#tdoc-share-close` so the published Playwright assertions work for both panels is a good choice.

Offline JUL-36 suite is green (20/20), including the new “no separate Share settings item + Share dispatches to manage” guard.

LGTM on the already-merged change. Two follow-ups that became more important once Share is the only entry:

**Follow-up:** After Sign out, `isOwner` is cleared but `cfg.ownerManage` is not. The old “Share settings” item lived in the identity menu and disappeared with the chip (`isOwner && cfg.ownerManage`). Share stays in the bar and now dispatches only on `cfg.ownerManage`, so a signed-out owner tab still opens Unpublish / Delete / the allowlist. Mutations still 401 (cookie is gone) — this is not a new write path — but the manage UI and allowlist remain visible. Clear `cfg.ownerManage` on logout, or gate the dispatch on the live `isOwner` flag the sign-out handler already updates.

**Nit:** Agree with Copilot — `showManageModal()` should call `closeManageModal()` before creating another panel (same pattern as `closeAuxModal` / `closeManageConfirm`). Re-clicking the always-visible Share button can stack duplicate `#tdoc-share-url` IDs. Pre-existing, more reachable now.

Playwright Share tests were updated but not re-run against a live worker; the offline source guards look right. Worth one owner-session click-through on the next publish.

---

PR #126, opened 2026-08-15
revert: tdoc.dev CD (#114) so Julie can land it

serenakeyitan:
I merged Julie's #114 by mistake. This puts `main` back to the pre-#114 tree so she can land CD herself.

Also reverts the follow-up #125 (Node 22 pin). That patch only exists because of #114.

@yayashuxue `feat/tdoc-dev-cd` is still on the remote. #114 itself cannot be reopened after a merge — please open a fresh PR from that branch when you want it.

Note from the accidental deploy: wrangler 4.90.1 needs Node 22. The first CD run died on Node 20 (`EBADENGINE`). #125 had that pin if you want to include it.

Does not undeploy the Worker that already went out to tdoc.dev. It only removes the workflow from `main`, so later merges will not auto-deploy until you land CD again.


yayashuxue:
Yes I probably will not merge 114 as it only supports single user tenant (current Tdoc.dev doesn't support multi user I just figured...)

serenakeyitan:
@yayashuxue tdoc.dev `TDOC_OWNER` is set back to `yayashuxue`. Sign in again and `/me` + Share should be yours. Sorry about the overwrite.

Issue #131, opened 2026-08-15
tdoc.dev/me should be the signed-in user's docs; localhost must publish there


serenakeyitan:
## What

`tdoc.dev` is supposed to be the hosted product, not Julie's personal worker with a pretty domain.

Two things are wrong together:

1. **`https://tdoc.dev/me` is Julie's catalog.** The Cloudflare Worker behind tdoc.dev is Julie's. `/me` lists every doc on **this worker**, and only `TDOC_OWNER` (Julie, signed in with GitHub) can see it. Anyone else — including a signed-in GitHub user who is not the owner — gets 302'd to `https://github.com/tornado-doc/tdoc`. GitHub sign-in today is for commenting on a single doc, not for "my docs."
2. **Other people's localhost cannot publish to tdoc.dev.** `/tdoc publish` from local studio still means first-time Cloudflare / Vercel self-host. Hosted registration on tdoc.dev is closed (`POST /api/hosted/token` 404s). So a normal user with a doc on `localhost:7878` has no way to push it to tdoc.dev.

If someone opens `tdoc.dev/me`, they should sign in with GitHub and see **their** published docs. If they hit Publish on localhost, that doc should land on tdoc.dev under that account. Julie should not see or own other people's docs on `/me`.

## Now (not a bug in the single-tenant code)

This is how a **one-owner worker** is designed:

- `/` never lists docs
- `/me` = docs hosted on this worker, owner-only
- everyone else publishes to their own `*.workers.dev` / Vercel

That is correct for self-host. It is the wrong product on `tdoc.dev`.

## Please solve

- **Hosted default.** Localhost Publish / `/tdoc publish` with no flag goes to `tdoc.dev`. `--platform cloudflare` and `--platform vercel` stay as self-host.
- **`/me` is per GitHub user**, not `TDOC_OWNER`'s dump of the whole worker. Sign in → your slugs. Sign out / not you → no enumeration of other people's docs (keep the current "no public catalog" rule).
- **Someone who is not Julie can publish from their own localhost** onto tdoc.dev and then find that doc on `/me`.

#109 already has the hosted publish target and a review on default + registration. Use that; don't start a parallel publish stack. The `/me` identity split is the missing product piece — once many accounts share one worker, today's owner catalog would list everyone's docs to Julie and nobody else.

## Not doing

- Do not make `/` a public directory of all docs
- Do not make self-host `/me` (a user's own worker) into a multi-tenant account page
[...]

PR #132, opened 2026-08-15
fix: Vercel publish hits public host, not SSO-gated deploy URL

serenakeyitan:
## Why

Clicking Publish on localhost showed `Failed: publish_failed`.

`tdoc-publish` uploaded to `published.json` `.base`, which was a per-deployment URL (`tdoc-….vercel.app`). That URL is behind Vercel Deployment Protection, so curl gets `401 Protected Deployment`. The production alias (`tdoc-phi.vercel.app`) is fine. The modal only printed `data.error`, so the real reason was hidden.

## What

- Vercel upload / pull / unpublish / agent-reply go through `public_host`
- First-time setup stores `base` as the public alias
- Publish modal appends the last stderr line

## Verify

- `node test/cli.test.js` — 10 passed
- Reproduced the 401 against the old `.base`, then published `tdoc-me` to https://tdoc-phi.vercel.app/d/tdoc-me/v/1

Not tdoc.dev / hosted (#131). This is the self-host Vercel path on this machine.

### DIFF
--- a/bin/tdoc-agent-reply
+++ b/bin/tdoc-agent-reply
@@ -110,7 +110,11 @@ if [ -f "$CONFIG_FILE" ]; then
   PLATFORM="$(jq -r '.platform // "cloudflare"' "$CONFIG_FILE")"
   TOKEN="$(jq -r '.upload_token // empty' "$CONFIG_FILE")"
   if [ "$PLATFORM" = "vercel" ]; then
+    HOST="$(jq -r '.public_host // empty' "$CONFIG_FILE")"
     BASE="$(jq -r '.base // empty' "$CONFIG_FILE")"
+    if [ -n "$HOST" ] && [ "$HOST" != "null" ]; then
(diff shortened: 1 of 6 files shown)


serenakeyitan:
Closing. Not worth landing in front of #109.

This PR only fixes the **self-host Vercel** upload URL (`.base` was a SSO-protected per-deploy alias). It does not send localhost Publish to tdoc.dev, and it does not make `/me` per-user.

Julie is rewriting the same files in #109 (`tdoc-publish` / `tdoc-pull` / `tdoc-unpublish` / `cli.test.js`). Merging this now is merge friction for the path we actually want: hosted default on tdoc.dev (#131).

Local workaround already applied on this machine (`published.json` `.base` → `tdoc-phi.vercel.app`). If Vercel self-host still needs the public-host fix after hosted lands, reopen or fold it into a follow-up. Not blocking #131.

PR #135, opened 2026-08-16
fix: stop clipping tables and diagrams in the reader

yayashuxue:
## Summary
- Overlay table styles used a `-14px` left margin that cropped the first column inside `overflow-x:auto` wrappers (the cf-vs-vercel matrix showed `CAPABILITY` as `APABILITY`).
- Keep real table layout (no `display:block` on `<table>`), scroll wide tables in a wrapper, and preserve SVG viewBox aspect so diagrams are not cropped.
- Reading column stays ~720px. This is a clip fix, not a wider page.

## Test plan
- [ ] `node test/run.js` (includes new `reader-overflow.test.js`)
- [ ] `node test/ui.test.js` (table not clipped on the left; SVG aspect preserved)
- [ ] Local smoke: serve `~/tdocs/cf-vs-vercel` with this overlay and confirm the capability column and architecture SVG are fully visible
- [ ] After tdoc.dev Worker deploy, reload https://tdoc.dev/d/cf-vs-vercel/v/1

### DIFF
--- a/test/ui.test.js
+++ b/test/ui.test.js
@@ -96,6 +96,63 @@ async function tPub(name, fn) {
     if (stored !== 'light') throw new Error(`storage should be light, got "${stored}"`);
   });
 
+  await t('Table in overflow-x:auto wrapper is not clipped on the left', async () => {
+    const m = await page.evaluate(() => {
+      const table = document.querySelector('.wrap table');
+      const wrap = table && table.parentElement;
+      const th = table && table.querySelector('th');
+      if (!table || !wrap || !th) return { missing: true };
+      const t = table.getBoundingClientRect();
+      const w = wrap.getBoundingClientRect();
+      const h = th.getBoundingClientRect();
+      return {
+        tableLeft: t.left,
+        wrapLeft: w.left,
+        thLeft: h.left,
+        thText: th.textContent.trim(),
+        tableMarginLeft: getComputedStyle(table).marginLeft,
+        wrapOverflowX: getComputedStyle(wrap).overflowX,
+      };
+    });
+    if (m.missing) throw new Error('fixture table/wrapper missing');
+    if (m.wrapOverflowX !== 'auto' && m.wrapOverflowX !== 'scroll') {
+      throw new Error(`expected overflow-x auto wrapper, got ${m.wrapOverflowX}`);
+    }
+    if (m.tableLeft < m.wrapLeft - 1) {
+      throw new Error(`table left ${m.tableLeft.toFixed(1)} is clipped inside wrapper ${m.wrapLeft.toFixed(1)}`);
+    }
+    if (m.thLeft < m.wrapLeft - 1) {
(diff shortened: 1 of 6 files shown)


test/ui.test.js
```diff
@@ -96,6 +96,63 @@ async function tPub(name, fn) {
     if (stored !== 'light') throw new Error(`storage should be light, got "${stored}"`);
   });
 
+  await t('Table in overflow-x:auto wrapper is not clipped on the left', async () => {
```
serenakeyitan, line 99: Optional follow-up: this asserts the original clip (author `overflow-x:auto` wrapper + no negative margin). `wrapScrollableTables()` returns early on that parent, so the new injected `.tdoc-table-scroll` path is not exercised here.

A second fixture table with no overflow wrapper (wide enough that `scrollWidth > clientWidth`) would lock the provider path: parent class `tdoc-table-scroll`, table not clipped, page not horizontally overflowing.

I smoke-tested that case against this overlay and it behaved correctly. Not blocking — the PR is already merged.

serenakeyitan (approved):

Already merged; posting the requested review.

This is the right fix. The `-14px` table pull was optical alignment that becomes a clip inside any `overflow-x:auto` wrapper, and `display:block` on `<table>` was the wrong mobile fallback. Keeping real table layout and scrolling a wrapper matches the AGENTS.md rule that reader invariants live in overlay code, not author HTML.

I checked the change against the Google review bar (design, functionality, tests, docs):

- Design is small and in the right layer. Overlay CSS/JS is the published reader, so this belongs here rather than in skill prompts.
- Functionality is good for readers. The 720px column stays put; this is a clip fix, not a wider page.
- Tests cover the original clip (author-wrapped table + no negative margin). Changelog is updated.

Verified on the PR overlay, not just by reading:

- `node test/run.js` — 29 offline suites green
- Headless Chrome, working-tree overlay:
  - sample-doc v2 (author `overflow-x:auto` wrap): first header is `Key`, `margin-left: 0`, `display: table` at 375px, not clipped
  - unwrapped wide table: gets `.tdoc-table-scroll`, table ~1094px inside a 672px scrollport, no page overflow, `CAPABILITY` fully visible
  - SVG `viewBox="0 0 800 200"` used 672×168 (`aspect-ratio: 800 / 200`), `overflow: visible`
  - reading column stays 720px
- Live https://tdoc.dev/d/cf-vs-vercel/v/1 is already on the post-merge overlay (`margin: 0 0 18px`, first column not clipped)

Optional follow-up (not a merge blocker):

The new `wrapScrollableTables()` path is the provider enforcement for docs that did not wrap the table, but `ui.test.js` only loads the already-wrapped fixture table, so that function is skipped. A Playwright case with a bare wide table would lock the wrapper + scrollport. I smoke-tested that path locally and it works.

Nit: Copilot's note on the `display:block` source guard is fair — the exact-string check is easy to miss with equivalent CSS.

LGTM.

Issue #142, opened 2026-08-16
Design: interactive onboarding page at /start


serenakeyitan:
## What

An interactive onboarding page at `/start`, reached from the landing page's "Create your first doc". Not a document to read: a page you act on, in the landing page's visual language, built from the same components.

Three steps, each with a one-click action rather than an instruction:

1. **Install** — copy the marketplace line (`/plugin marketplace add tornado-doc/tdoc`) or the ONBOARDING.md paste line, with the runtime logos showing it works with the agent they already have.
2. **Make something** — no CLI to memorise. The page shows what to say to the agent in plain words, with a few prompts they can copy as-is.
3. **Learn the loop on a real doc** — the part that matters, below.

## The loop, taught on the real Conway doc

The user forks `conway-life` (the published "What if a doc could think?"), which gives them a doc of their own with the comments already on it. Then, in their own copy:

1. Select a comment on the board.
2. Ask their agent to fix it.
3. A new version appears in the version picker.
4. Open it: the change is applied, and the agent has replied on the thread under its own identity.

That is the whole product in four actions, on real content, with no writing required from the user.

## Hosting

Default is tdoc.dev, with switching to your own Cloudflare or Vercel presented as something you ask the agent for, not a command to look up.

**This part is blocked, see below.**

## ⚠️ Blocker: the default host does not exist

`bin/tdoc-publish` deploys to the user's **own** Cloudflare account and builds `PUBLIC_HOST="tdoc.${SUBDOMAIN}.workers.dev"`. When it cannot find a claimed workers.dev subdomain it aborts rather than guessing. There is no tdoc.dev-hosted publish path today.

So "hosts on tdoc.dev by default, switch to your own if you want" describes a product that has not been built. Onboarding cannot teach it. One of:

- **A.** Build the hosted default first (multi-tenant publish on tdoc.dev, auth, quotas, abuse story). Large, and it is a product decision, not a docs decision.
- **B.** Ship onboarding against what exists: you publish to an account you own, and the page walks the two browser clicks Cloudflare needs.
- **C.** Ship onboarding without a publish step at all: fork, comment, fix, reply, all on tdoc.dev's copy of the Conway doc, and leave publishing to a later page.

**C** is the smallest thing that teaches the loop, since the fork flow already runs on tdoc.dev and needs no account from the user. Recommending C unless the hosted default is already planned.

## A second constraint worth knowing

[...]

Issue #138, opened 2026-08-16
Author JS is an embedded-artifact capability, not a document capability


serenakeyitan:
## Decision

**Arbitrary JavaScript is not a document capability. It is an embedded-artifact capability.**

The published page (prose, images, tables, comments) stays on the trusted origin and keeps today's CSP: author `<script>` / `onclick` stay inert. That is still correct — see Why it's closed below.

Anything that actually has to compute (Game of Life, a live slider, a calculator) is a **sandboxed interactive island**: its own HTML resource, in an `<iframe sandbox="allow-scripts">` with **no** `allow-same-origin`. The overlay comments on the iframe as one artifact (`iframe[src]` is already in the commentable set). It does not walk into the frame.

```
trusted document
├── prose / images / SVG / tables
├── iframe: untrusted interactive artifact
│     └── arbitrary JS, nested iframes, whatever
└── comments overlay (outside the frame)
```

tdoc does not care what happens inside the island. The island cannot read the session cookie or call `DELETE /api/doc` / `PATCH /api/doc/access` with credentials. Worst case a malicious island is an ugly box; it cannot delete the doc.

This replaces the earlier "put the whole document in an iframe and split the overlay across the frame boundary" idea. That path would rebuild comment anchoring — the product — to buy script support. Too much. Do not do it.

## Why the document itself stays closed

`cspHeader()` (`worker/worker.js`, mirrored in `server/server.js`) serves every doc under:

```
script-src 'nonce-<overlay>' 'strict-dynamic'; object-src 'none'; base-uri 'none';
```

Only the overlay's injected tags carry the nonce. Author `<script>`, inline `onclick=`, and `javascript:` URLs are inert — locally and published. `test/csp-headers.test.js` asserts this.

The reason is a real confused-deputy attack, not caution. Doc HTML is **same-origin with the control plane**, and owner mutations (`DELETE /api/doc`, `PATCH /api/doc/access`) can be authorized by the owner's **session cookie alone**. A `<script>` in the doc could `fetch('/api/doc', {method:'DELETE'})`, the browser would attach the cookie, and the owner's docs would be gone with no gesture. Closing that was correct — #63, #68.

Giving the *document* "a little JS" and then trying to forbid the bad cases is a recursive mess (child scripts, nested iframes, workers, fetch, what the overlay should follow). The island boundary is what stops that recursion.

## Why this is smaller than #138's old option 2

[...]

PR #143, opened 2026-08-16
Interactive onboarding page at /start

serenakeyitan:
Related to #142

Builds the onboarding page the homepage CTA points at. Three steps, each one an action rather than an instruction.

**1. Install** — pick your runtime (Claude Code, Codex, Cursor, Gemini, Grok), get the marketplace line.

**2. Say what you want** — two prompts you can lift as-is. Hosting is folded behind a disclosure, and every target shows what to *say*, never a command to type. A deploy command in that panel would contradict the page it sits on.

**3. The loop** — on the real published Conway doc, not a mockup: fork it, ask your AI to fix a comment, a new version appears, the agent replies under its own identity.

### No author JavaScript

Published tdocs run under a nonce-only CSP, so both pickers are radio inputs driving sibling selectors and the disclosure is `<details>`. A page that needed script would look right while authoring and be dead once shared, which is exactly #138. A test asserts no `<script>`, no inline handlers, and that the pickers are `:checked`-driven.

### `/start` fails safe

Routes through `landingResponse` with the slug parameterised, inheriting the same fallback as `/`: unpublished, access-gated, or throwing all degrade to the neutral page rather than 404. Only tdoc.dev has the doc, so self-hosted workers are unchanged with no configuration.

### One thing to know before reviewing

The hosting step is written **as though publishing to tdoc.dev by default already works**. It does not: `bin/tdoc-publish` deploys to an account you own and builds `tdoc.<subdomain>.workers.dev`. That was the owner's explicit call, to design the flow against the finished product rather than block on it. The page will describe reality once the hosted default ships.

Also: `/fork` exists in `worker/worker.js` but not `server/server.js`, so step 3's flow can only be walked against a deployed worker.

Tests: 29/29 offline suites green, including 6 new assertions for this page.

**Not merging.**


yayashuxue:
Closing as superseded by #153 (`Onboarding: one line, no account`), which landed on `main` and explicitly replaces this PR. Issue #142 is closed. `/start` and the `tdoc-start` doc are already on `main`.

This branch was also written against the pre-#129 landing tree (51 files, including a stack of `landing/tornado-doc` versions). Rebasing it would fight the page #153 already shipped.

serenakeyitan:
Agreed. This PR stays closed.

Checked the close against `main` and production:

- #153 is merged and already says it replaces this PR.
- `/start` is on `main` (`START_SLUG = 'tdoc-start'`, fail-safe `landingResponse`) and live at https://tdoc.dev/start — 200, title is "tdoc, the full tutorial", not this branch's three-step page.
- `landing/tdoc-start` and the `/start` tests on `main` explicitly reject the runtime picker / `/plugin marketplace` flow this branch shipped.
- #142 is closed as completed.

This branch is the pre-#129 landing tree (51 files, `tornado-doc` v1–v36). Rebasing it would fight the page that already shipped. Leftover first-doc work is #157 / #158, not this stack.

---

PR #145, opened 2026-08-16
fix(auth): org OAuth SoT + landing redirects

yayashuxue:
## Summary
- Switch shared Device Flow to the **tornado-doc** OAuth App (`Ov23li1jJiBkS23x4O07`) via `shared/github-oauth.js` (single SoT; template + Vercel consume it; BYOK backfill retires Serena’s personal client id).
- Unauthorized `/me` and unknown GET paths redirect to **landing** with a toast (`?notice=…`), not github.com; landing can sign in with GitHub.
- Soft-land OAuth approve on `/auth/done` (set the OAuth App callback URL here to avoid GitHub 404 after Authorize).
- After Device Flow succeeds, refresh `isOwner` from `/api/auth/me` so **My docs** appears in the profile menu without a reload.

## Test plan
- [ ] `npm test` (31 offline suites)
- [ ] Sign in on a published doc → profile menu shows **My docs** when `TDOC_OWNER` matches (no reload)
- [ ] Visit `/me` signed out → lands on `/?notice=signin` with toast + Sign in
- [ ] Visit `/me` signed in as non-owner → `/?notice=me` toast
- [ ] Visit `/auth/done` → “You’re signed in / close this tab”
- [ ] Confirm org OAuth App callback URL is `https://<host>/auth/done` (e.g. `https://tdoc.dev/auth/done`)
- [ ] Redeploy BYOK worker (or next publish that redeploys) picks up org client id if wrangler still had the old personal id

### DIFF
--- a/bin/tdoc-publish
+++ b/bin/tdoc-publish
@@ -299,7 +299,7 @@ vercel_bundle_sha() {
     const crypto = require("crypto");
     const skill = process.env.SKILL_DIR;
     const h = crypto.createHash("sha256");
-    for (const rel of ["worker/worker.js", "server/overlay.js"]) {
+    for (const rel of ["worker/worker.js", "server/overlay.js", "shared/github-oauth.js"]) {
       h.update(rel); h.update("\0"); h.update(fs.readFileSync(path.join(skill, rel))); h.update("\0");
     }
     const dir = path.join(skill, "vercel");
@@ -424,6 +424,8 @@ first_time_setup() {
   else
     echo "[tdoc] no GitHub login detected — doc catalog stays private (link-only)."
   fi
+  # GITHUB_CLIENT_ID comes from wrangler.toml.template (kept in sync with
+  # shared/github-oauth.js by test/no-drift.test.js).
   sed \
     -e "s/PLACEHOLDER_KV_ID/${kv_id}/g" \
     -e "s/PLACEHOLDER_OWNER/${owner_login}/g" \
@@ -485,6 +487,9 @@ sync_vercel_app() {
   mkdir -p "$VERCEL_APP_DIR/api" "$VERCEL_APP_DIR/lib"
   cp "$SKILL_DIR/vercel/api/"*.js "$VERCEL_APP_DIR/api/"
   cp "$SKILL_DIR/vercel/lib/"*.js "$VERCEL_APP_DIR/lib/"
+  # Flat-copy SoT so api/tdoc.js can require('../lib/github-oauth.js') without
+  # a ../../shared hop (vercel-app is outside the skill checkout).
+  cp "$SKILL_DIR/shared/github-oauth.js" "$VERCEL_APP_DIR/lib/github-oauth.js"
   cp "$SKILL_DIR/vercel/vercel.json" "$SKILL_DIR/vercel/package.json" "$VERCEL_APP_DIR/"
   SKILL_DIR="$SKILL_DIR" bundle_worker "$VERCEL_APP_DIR"
 }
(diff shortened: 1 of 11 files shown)


worker/worker.js
```diff
@@ -977,8 +977,17 @@ function injectOverlay(rawHtml, slug, version, identity, versions, isOwner, owne
...
+// brand + sign-in (when auth is configured) + a link to the open-source
+// project. Docs are link-only. `notice` is an optional toast reason when
+// we bounce users here from /me or an unknown path.
+function landingHtml(env, notice) {
+  const authOk = !!(env && String(env.GITHUB_CLIENT_ID || '').trim());
+  const toastMsg = ({
+    me: 'My docs is only available after you sign in as the worker owner.',
+    signin: 'Sign in with GitHub to continue.',
+    notfound: 'That page was not found. Sign in or open a doc from its shared link.',
+  })[notice] || '';
```
serenakeyitan, line 1023: Nit: this lookup is not an allowlist. `({me,signin,notfound})['constructor']` is `Object` (truthy), `JSON.stringify(Function)` is `undefined`, and the inline script becomes `var toastMsg = ;` — a SyntaxError that disables Sign in on `/?notice=constructor`.

```js
const TOASTS = { me: '...', signin: '...', notfound: '...' };
const toastMsg = Object.hasOwn(TOASTS, notice) ? TOASTS[notice] : '';
```

bin/tdoc-publish
```diff
@@ -299,7 +299,7 @@ vercel_bundle_sha() {
     const crypto = require("crypto");
     const skill = process.env.SKILL_DIR;
     const h = crypto.createHash("sha256");
-    for (const rel of ["worker/worker.js", "server/overlay.js"]) {
+    for (const rel of ["worker/worker.js", "server/overlay.js", "shared/github-oauth.js"]) {
```
serenakeyitan, line 302: Should: Vercel now hashes `shared/github-oauth.js`, but `worker_bundle_sha()` above still only hashes `worker.js` + `overlay.js`.

This PR still redeploys Cloudflare BYOK because overlay.js changed, so the personal-id backfill will run once. A later ID-only rotation would update the template and skip the CF redeploy/backfill.

Please add the SoT file (or `wrangler.toml.template`) to `worker_bundle_sha()` so the hash actually tracks the client id.

serenakeyitan (approved):
LGTM with comments. Net improvement: org OAuth as a single SoT, `/me` and unknown browser paths stay on tdoc instead of bouncing to github.com, `/auth/done` stops the post-Approve 404, and Device Flow now refreshes `isOwner` so My docs appears without a reload.

`npm test` is green here (31/31 offline). CI offline suite + CodeQL are also green.

I agree with Copilot on the two inline notes already filed: drop the empty `for` in `test/no-drift.test.js`, and do not 302 unknown `GET /api/*` (browser `fetch()` follows redirects and then `r.json()` eats HTML). Please take those.

Two more non-blocking notes inline. Also:

- Add an Unreleased CHANGELOG line for the org OAuth switch and the `/me` landing bounce.
- Follow-up, not this PR: signing in as owner still leaves `ownerManage` null until reload, so Share stays copy-link-only until refresh. The stated My docs fix is correct.

Ops before/with deploy (not code):
1. After this is on tdoc.dev, set the tornado-doc OAuth App callback to `https://tdoc.dev/auth/done`. One callback URL per app. BYOK still completes via Device Flow poll on the original tab; the GitHub tab will land on tdoc.dev, which is fine.
2. Existing BYOK Cloudflare workers pick up the new client id on the next publish that redeploys. This PR also touches overlay.js, so that hash will change once.

PR #163, opened 2026-08-18
Download PDF uses browser print, not a blurry JPEG

yayashuxue:
## Summary
- **Download PDF** now prints the `/export` reading column (reader CSS, no bar) so **Save as PDF** keeps real vector text.
- Removes the ~100 DPI canvas JPEG wrap that made PDFs look mushy.
- HTML download is unchanged (`slug-vN.html`).

## Test plan
- [ ] Open a published doc, **Download → Download PDF** — print dialog appears, not a silent `.pdf` file
- [ ] Destination **Save as PDF**; text stays sharp when zoomed
- [ ] Print output has the reading column only (no tdoc bar / comments)
- [ ] **Download HTML** still saves `slug-vN.html`
- [ ] `npm test`

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -913,7 +913,8 @@
     paintTheme(next);
   };
 
-  // Duplicate = hosted account copy. Download = HTML file or PDF snapshot.
+  // Duplicate = hosted account copy. Download HTML = /export file.
+  // Download PDF = print the export (browser Save as PDF), not a JPEG wrap.
   let pendingDuplicate = false;
   function downloadExport() {
     const a = document.createElement('a');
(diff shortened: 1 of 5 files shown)


server/overlay.js
```diff
@@ -1077,43 +951,18 @@
...
-      document.body.appendChild(a);
-      a.click();
-      a.remove();
-      setTimeout(() => URL.revokeObjectURL(blobUrl), 4000);
-    } finally {
-      iframe.remove();
+      win.addEventListener('afterprint', drop, { once: true });
+      setTimeout(drop, 120000);
+      win.focus();
+      win.print();
```
serenakeyitan, line 966: `win.print()` runs only after iframe load + image decode + two rAFs. Desktop Chrome still opens the dialog; Safari/Firefox may ignore `print()` once the click gesture is gone, and this function does not throw — so `startDownload` never shows “Could not download PDF”.

Please verify Safari once. If it no-ops, open `/export` in a window during the click and print that document.

Not blocking if Safari works.

serenakeyitan (approved):
<h3>Breeze Summary</h3>

Replaces the canvas JPEG wrap + hand-rolled PDF with `contentWindow.print()` on `/export?download=0`, so Save as PDF keeps real vector text. Download HTML is unchanged.

- HTML still uses `/export?download=1` and `slug-vN.html`
- PDF iframe still uses `/export?download=0` (inline, reader CSS, no bar)
- `jpegPagesToPdf` / `toDataURL('image/jpeg')` are gone
- Reader CSS still has `@media print` and `@page`
- `npm test`: 34/34 offline suites green; CI CodeQL + offline suite green
- Headless Chrome: 20/20 times this iframe `onload` pattern resolved to the export document, not `about:blank`

Safe to merge. Not 5 because I could not click **Download PDF** on a published worker, and `print()` after the iframe await can fail with no dialog and no error modal.

| Filename | Overview |
|---|---|
| `server/overlay.js` | Deletes JPEG PDF; prints the export iframe |
| `test/duplicate-download.test.js` | Flips the old “file download, not print” contract |
| `CHANGELOG.md` / `SKILL.md` | Documents print / Save as PDF |

<h3>Issues</h3>

1. **`server/overlay.js` — `print()` after awaits can be ignored.** `downloadPdf()` awaits iframe load, `img.decode()`, and two rAFs, then calls `win.print()`. Desktop Chrome allows this; Safari/Firefox may swallow it with no exception, so the existing error modal never appears. Please click the menu once in Safari. If the dialog does not open, open `/export` during the click (user gesture) and print that window.

2. **`server/overlay.js` — invisible on-screen iframe + `win.focus()`.** Agree with Copilot: `left:0;top:0` plus focus can trap keyboard users on an `opacity:0` iframe until `afterprint` or the 120s timeout. The JPEG path used `left:-12000px` and did not steal focus. Keep it off-screen, `aria-hidden`, and restore focus on `drop()`.

3. **`server/overlay.js` — `afterprint` removes the iframe immediately.** If preview still needs the document, the PDF can go blank. Delay `drop()` a few seconds after `afterprint`, and clear the 20s load timer (Copilot’s second comment).

4. **`server/overlay.js` — label still says Download PDF.** Intentional reversal of `de729a3` (“must be a file download, not the print dialog”). Fine for sharpness; consider “Print / Save as PDF” so the dialog is not a surprise.

No blocking issues.

---

PR #181, opened 2026-08-20
fix: notification deep-link reveals the target comment/reply

### LINKED ISSUE #180: Notifications don't redirect to the target doc/comment
Clicking a notification (new comment / mention) does not appear to take the user to the relevant doc or the specific comment.

### Expected
- Notification click → opens the target doc.
- If the notification is about a specific comment, scroll to and highlight that comment.

### Steps to reproduce
1. Trigger a notification (add a comment / @mention on a doc).
2. Click the notification.
3. Observe it does not land on the doc/comment.

cc @serenakeyitan

yayashuxue:
Fixes #180.

## Problem
Clicking a notification takes the reader to the doc but **not** to the comment/reply it was about — you land at the top of the doc with the referenced thread collapsed.

## Root cause
The deep-link handler in `refreshComments` (overlay.js) resolved its target and added the thread to `state.openReplyThreads` inside a **post-layout `requestAnimationFrame`** — i.e. *after* `buildCard()` had already rendered every card. A card only renders its reply list expanded when its id is in `openReplyThreads` **at build time**. So for `reply` and `reaction` notifications (whose target id is a reply), the thread rendered collapsed and the referenced reply was never revealed or scrolled to.

The same-doc click path in `openInboxTarget` had the same ordering flaw and additionally never triggered a re-render, so opening the thread had no visible effect at all.

## Fix
- Resolve the deep-link target **before** the card-building loop and open its thread up front, so the reply element is in the DOM and expanded when we scroll to it.
- Route the same-doc notification click through `refreshComments` via a new `state.pendingDeepLink`, so it uses the same corrected path: open → activate → scroll → mark-seen.

Single source of truth: `server/overlay.js` also fills the worker bundle (`__TDOC_OVERLAY_JS__`), so this covers both local and hosted.

## Testing
- `notifications.test.js` (9), `overlay-pure.test.js` (15), `browser-bundles-parse.test.js` (5) all pass.
- `node --check server/overlay.js` clean.

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -765,6 +765,7 @@
     pinnedId: null,                // comment whose floating card is click-pinned open (wide mode)
     hoverId: null,                 // comment whose card is open via hover (wide mode)
     openReplyThreads: new Set(),   // top-level comment ids whose reply lists stay expanded
+    pendingDeepLink: null,         // comment/reply id a notification click wants revealed on next refresh
   };
 
   // Highlight API: one shared registry for pending, one per saved comment.
@@ -1284,14 +1285,13 @@
     const destSlug = row.slug || slug;
     const destVer = row.version || version;
     if (destSlug === slug && Number(destVer) === Number(version)) {
-      if (root) {
-        state.openReplyThreads.add(root);
-        pinOpenCard(root);
-        if (target === root) setActiveComment(root);
-      }
-      const el = document.querySelector(`[data-comment-id="${CSS.escape(target || '')}"]`);
-      if (el && el.scrollIntoView) el.scrollIntoView({ block: 'center' });
-      markInboxSeen(target);
+      // Route through refreshComments so the target thread is opened BEFORE its
+      // card is (re)built and the reply/reaction target is actually revealed.
+      // Poking the DOM here instead raced the collapsed cards and left replies
+      // hidden. refreshComments handles opening, activating, scrolling, and
+      // markInboxSeen for the pending target.
+      state.pendingDeepLink = target || root || null;
+      refreshComments();
       return;
     }
     location.href = `/d/${encodeURIComponent(destSlug)}/v/${destVer}?comment=${encodeURIComponent(target || root || '')}`;
@@ -2639,6 +2639,24 @@
     document.body.classList.toggle('tdoc-has-comments', state.activeComments.length > 0);
     document.body.dataset.tdocReady = '1';
 
+    // Resolve the deep-link target (notification click or ?comment= on load)
+    // BEFORE building cards. A card only renders its reply list expanded when its
+    // id is in state.openReplyThreads at build time, so a reply/reaction target
+    // must open its thread up front — otherwise buildCard renders it collapsed and
+    // the referenced reply is never revealed or scrolled to. `want` is consumed in
+    // the post-layout rAF below.
+    const want = allowDeepLink
+      ? (state.pendingDeepLink
+          || (() => { try { return new URLSearchParams(location.search).get('comment'); } catch { return null; } })())
+      : null;
+    let deepLinkRoot = null;
+    if (want) {
+      const hit = state.activeComments.find(c => c.id === want)
+        || state.activeComments.find(c => (c.replies || []).some(r => r.id === want));
+      deepLinkRoot = hit ? hit.id : want;
+      state.openReplyThreads.add(deepLinkRoot);
+    }
+
     const fabCount = document.getElementById('tdoc-fab-count');
     if (fabCount) fabCount.textContent = state.activeComments.length;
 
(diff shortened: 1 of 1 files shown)


server/overlay.js
```diff
@@ -2713,21 +2731,18 @@
         // and lost the card's .active state (+ the "move anchor" affordance).
         setActiveComment(keepPinnedId);
```
serenakeyitan, line 2732: This `setActiveComment(keepPinnedId)` now runs on the same-doc inbox-click path, because that path was rerouted through `refreshComments()`. `setActiveComment` always calls `markInboxSeen(id)`.

If the reader already has a card pinned (typical when they click a notification on the doc they are looking at), this marks the **pinned** comment as read before `if (want)` runs. When `want` is a reply, that is the root (or some other thread) — the side effect the comment a few lines below says to avoid.

Skip the restore when a deep-link is in play:

```js
if (keepPinnedId && !want && !state.narrow && state.cardEls.has(keepPinnedId)) {
  setActiveComment(keepPinnedId);
}
```

The `if (want)` branch already pins/activates the right target.

server/overlay.js
```diff
@@ -2639,6 +2639,24 @@
...
+    // Resolve the deep-link target (notification click or ?comment= on load)
+    // BEFORE building cards. A card only renders its reply list expanded when its
+    // id is in state.openReplyThreads at build time, so a reply/reaction target
+    // must open its thread up front — otherwise buildCard renders it collapsed and
+    // the referenced reply is never revealed or scrolled to. `want` is consumed in
+    // the post-layout rAF below.
+    const want = allowDeepLink
+      ? (state.pendingDeepLink
+          || (() => { try { return new URLSearchParams(location.search).get('comment'); } catch { return null; } })())
+      : null;
```
serenakeyitan, line 2651: `allowDeepLink === false` (inbox poll: `refreshComments({ deepLink: false })`) currently drops `state.pendingDeepLink` as well as `?comment=`.

Poll should not re-scroll to the URL param on every inbox tick. It should still honor an explicit notification click that already set `pendingDeepLink`. Otherwise a poll that interleaves with the click's in-flight fetch neither opens the thread nor clears the pending id, and a later default `refreshComments()` can jump/mark-seen at a surprising time.

```js
const want = state.pendingDeepLink
  || (allowDeepLink
      ? (() => { try { return new URLSearchParams(location.search).get('comment'); } catch { return null; } })()
      : null);
```

serenakeyitan (changes requested):
<h3>Breeze Summary</h3>
Fixes #180 by resolving the notification/`?comment=` target **before** `buildCard()`, so `openReplyThreads` is populated at render time and reply/reaction targets are not stuck in `.tdoc-replies { display:none }`. Same-doc inbox clicks now go through `state.pendingDeepLink` + `refreshComments()` instead of poking an already-built collapsed DOM.

Verified:
- Local fixture `?comment=r_fixture_1` at 1400px pins mode: `.tdoc-replies.open`, card `tdoc-floating-open`, reply height > 0, root **not** `.active`.
- `?comment=c_fixture_1`: root card is `.active.tdoc-floating-open`.
- No query param: replies stay collapsed, card `display:none` in pins mode.
- Narrow headless default: reply deep-link still opens the thread.
- `notifications.test.js` / `overlay-pure.test.js` / `browser-bundles-parse.test.js` pass, but none of them exercise this overlay path.
- Offline CI + CodeQL are green.

<h3>Confidence Score: 3/5</h3>
Do not merge yet. Routing same-doc clicks through `refreshComments()` now runs the `keepPinnedId` restore, which `setActiveComment`s the previously pinned card and `markInboxSeen`s it — the unread-state leak this PR's own comment says to avoid. Score is not 5 because that path is code-traced (inbox needs auth) while only the `?comment=` happy path was browser-checked.

| Filename | Overview |
|---|---|
| `server/overlay.js` | Pre-open deep-link thread before card build; `pendingDeepLink` for same-doc inbox clicks |

<h3>Issues</h3>

1. **`server/overlay.js` — Deep-link loses to `keepPinnedId` restore (blocking)**  
   Same-doc inbox clicks now call `refreshComments()`, so the post-layout rAF still does `setActiveComment(keepPinnedId)` *before* the `if (want)` branch. `setActiveComment` always `markInboxSeen`s that id. If a card is already pinned (the common case on this path) a reply notification will mark the root — or a different pinned comment — as read, which is exactly what the new comment says must not happen. Skip the restore when `want` is set: `if (keepPinnedId && !want && …)`.

2. **`server/overlay.js` — `pendingDeepLink` is gated by `allowDeepLink` (blocking)**  
   Agree with Copilot on this one. Inbox poll uses `refreshComments({ deepLink: false })`. That should skip the URL `?comment=` replay, not an explicit notification click. As written, a poll that wins the in-flight race ignores and does not consume `pendingDeepLink`, so a later default refresh can jump/mark-seen unexpectedly. Read `pendingDeepLink` even when `allowDeepLink` is false.

3. **none — No test covers the ordering bug this PR fixes**  
   The tests listed in the PR body never touch overlay deep-link behavior. `test/fixtures/tdocs/sample-doc/comments.json` already has `r_fixture_1`. Add a `ui.test.js` case: load `?comment=r_fixture_1`, assert `.tdoc-replies.open` and the reply's box height is > 0; without the query param, replies stay collapsed. That is the regression that shipped once already.

4. **none — Nit: Unreleased changelog**  
   User-facing inbox behavior. A one-line `Fixed` bullet under Unreleased would match how other overlay fixes are recorded.

---

serenakeyitan:
Superseded — closing.

This PR is `Fixes #180`, but **#180 was already fixed and closed on 08-20 by the merged PR #182** ("Notification clicks open the target doc and comment"), which addressed both causes: the same-doc click bubbling to the `click-outside` handler, and `/me` and `/` having no reliable path to the target comment.

This branch is also `CHANGES_REQUESTED` and currently conflicting. If there is behaviour here that #182 did not cover, a fresh issue against current `main` would be the cleaner path.

Issue #179, opened 2026-08-20
oauth path not native


serenakeyitan:
### What problem does this solve?

after the GitHub login page popup, it cannot redirect back to tdoc. it should always open a new tab for GitHub tab. not change the existing url of the tdoc page.

PR #184, opened 2026-08-20
Sign-in: put the device code on the clipboard, not just on screen

yayashuxue:
Builds on #183 (which gave sign-in a reliable Open-GitHub tab). Rebased onto it — this PR is now just the one thing #183 left out.

## The remaining gap

#183 fixed the *new tab*. But the code is still **copy-only** — it's only ever shown, never placed on the clipboard. Many visitors hit GitHub's **login page first**, which does not display the device code. In the branch where GitHub returns only `verification_uri` (no pre-fill), the visitor has to *paste* the code — so they go **Back** to grab it and lose the auth tab. That's the exact miss the flow exists to prevent.

## The fix

**Auto-copy the code the moment it exists**, layered on #183's design:

- **Desktop:** the click that opened the modal still counts as recent user activation, so the write lands silently — the Copy button reads **"Copied"** with no action needed.
- **iOS Safari** (activation dropped across the `device/start` round-trip): the auto-write is rejected and we stay quiet — no false "Copied", no misleading "Press ⌘C". The **"Open GitHub" tap** then copies the code from a real gesture, so it's on the clipboard when GitHub asks.

`verification_uri_complete` still pre-fills the code, so it's covered three ways: **pre-filled + auto-copied + copied on the tap that leaves.**

Shared module → the overlay comment sign-in, the landing page, and onboarding all get it.

## Verification

- All 40 default test suites green; `tds-copy`, the `Press ⌘C`/`Ctrl+C` fallback, `.tds-bg`, `.tds-open`, and the `tdoc:signedin` event are all preserved.
- Drove both paths in real Chrome against a stubbed device API:
  - **Desktop:** code auto-copies on open (`clipboard = [WDJB-MJHT]`, button "Copied"); `.tds-open` anchor carries the pre-filled URL, `target=_blank`.
  - **iOS Safari:** auto-write rejected → clipboard empty, button stays "Copy"; the **Open-GitHub tap copies the code** (`clipboard = [WDJB-MJHT]`).
- Also fixed two stale selectors in the gated `test/ui.test.js` (`#tdoc-device-modal` → `.tds-bg`, `#tdoc-modal-cancel` → `#tds-cancel`) that predate the shared sign-in module.

### DIFF
--- a/server/signin.js
+++ b/server/signin.js
@@ -118,8 +118,9 @@
         var code = document.getElementById('tds-code');
         code.textContent = d.user_code;
         var cbtn = document.getElementById('tds-copy');
+        var copied = false;
+        function mark() { copied = true; cbtn.textContent = 'Copied'; cbtn.className = 'tds-copy done'; }
         function copyCode() {
-          var mark = function () { cbtn.textContent = 'Copied'; cbtn.className = 'tds-copy done'; };
           if (navigator.clipboard && navigator.clipboard.writeText) {
             navigator.clipboard.writeText(d.user_code).then(mark, function () { selectCode(); });
           } else { selectCode(); }
@@ -135,6 +136,21 @@
         }
         cbtn.onclick = copyCode;
         code.onclick = copyCode;
+
+        // Put the code on the clipboard the moment it exists, so a visitor who
+        // lands on GitHub's login first (which never shows the code) can paste
+        // it without coming back for it — the miss this whole flow is built to
+        // avoid. On desktop the click that opened this modal still counts as
+        // recent activation, so this lands silently and the button reads
+        // "Copied". Where activation is gone (Safari drops it across the
+        // device/start round-trip) it stays quiet: no false "Copied", and the
+        // Open-GitHub tap below copies again from a real gesture. Never fall
+        // back to the "Press Ctrl+C" hint here — that's only honest after a
+        // real click.
+        if (navigator.clipboard && navigator.clipboard.writeText) {
+          navigator.clipboard.writeText(d.user_code).then(mark, function () {});
+        }
+
         var uri = d.verification_uri_complete || d.verification_uri;
         var where = document.getElementById('tds-where');
         // A native target=_blank anchor is the only hop that every browser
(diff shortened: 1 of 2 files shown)


serenakeyitan (approved):
<h3>Breeze Summary</h3>
Follow-up to #183: the shared device-flow dialog now puts `user_code` on the clipboard as soon as it exists (silent no-op if the browser rejects the write), and the Open GitHub tap copies again from a real gesture so a visitor who lands on GitHub's login page can paste without coming back. Overlay, landing, and onboarding all get it via `server/signin.js`.

Verified:
- Auto-copy uses `writeText(...).then(mark, function () {})` — no false "Copied", no "Press ⌘C" on a background reject. Open GitHub still copies only when `copied` is false.
- Chrome desktop stub: Sign-in click auto-copies `WDJB-MJHT`, Copy button reads **Copied**, `.tds-open` is `target=_blank` + `noopener noreferrer` + `verification_uri_complete`.
- iOS simulation (first `writeText` rejected): button stays **Copy**; the Open GitHub tap writes `WDJB-MJHT` and flips to **Copied**; this tab stays; GitHub opens at login with `user_code` in `return_to`.
- `signin-github-tab.test.js`, `tdoc-start.test.js`, `browser-bundles-parse.test.js` pass. `#tdoc-device-modal` / `#tdoc-modal-cancel` are gone; gated `ui.test.js` now waits on `.tds-bg` / `#tds-cancel`.
- CI offline suite + CodeQL are green.

Merge. Score is not 5 because the new clipboard invariant is not locked in the offline suite (the #179 tab tests still pass if this copy path is deleted), and the iOS path was simulated in Chrome rather than driven on Safari.

| Filename | Overview |
|---|---|
| `server/signin.js` | Auto-copy on code receipt; Open GitHub tap copies if the auto-write did not land |
| `test/ui.test.js` | Gated published sign-in selectors updated to the shared `.tds-bg` / `#tds-cancel` dialog |

<h3>Issues</h3>
No blocking issues.

1. **`test/signin-github-tab.test.js` — No source guard for the clipboard belts (nit)**  
   The existing tests only assert the `#179` tab contract. A couple of string checks (auto-copy `writeText` with an empty reject, Open GitHub click copies when `!copied`, auto-copy before `window.open`) would make this un-deleteable the same way the tab hop is.

2. **`server/signin.js` — Open GitHub handler calls `copyCode()` (nit)**  
   `copyCode()`'s failure path is `selectCode()` ("Press ⌘C"). If auto-copy is still in flight when the tap happens, a later reject can overwrite a successful "Copied". Same silent write as the auto-copy (`then(mark, function () {})`) avoids that without changing the iOS belt.

3. **`CHANGELOG.md` — Unreleased note (nit)**  
   User-facing sign-in behavior. A one-line `Fixed` bullet under Unreleased would match how other overlay/sign-in fixes are recorded.

---

yayashuxue:
🌐 **Live preview:** https://pr-184-tdoc-preview.jyshi1107.workers.dev — click **Sign in with GitHub** to exercise the modal (seeded doc: [/d/conway-life/v/2](https://pr-184-tdoc-preview.jyshi1107.workers.dev/d/conway-life/v/2)).

Deployed with `wrangler versions upload --preview-alias pr-184` to the isolated `tdoc-preview` Worker (own KV + R2, `TDOC_PREVIEW=1`, no Durable Object) — does not touch the production `tdoc.dev` deployment. Verified end-to-end: real GitHub device flow renders, code auto-copies on a real desktop click, and the Safari path falls back to copy-on-tap.

PR #186, opened 2026-08-20
Sign-in: let the visitor open GitHub, never open it for them

yayashuxue:
Follow-up to #183/#184. Fixes the "why did it just throw me to GitHub?" confusion.

## The problem

#179 gave the sign-in dialog a real **Open GitHub** anchor *and* an auto `window.open()` "convenience popup" beside it. On desktop that popup **lands**, so the dialog yanks you into a GitHub tab the instant it appears — before you've read the code or understood what's happening. It reads as a hijack, not a sign-in. (Reported first-hand: "自动跳转我都是蒙蔽的" — even with the code auto-copied, the auto-jump is disorienting.)

## The fix

**Drop the auto-open. The anchor is the only hop, and the visitor takes it.**

- Until they tap, the dialog just sits there: code already on the clipboard ("Copied"), and a clear **"Click Open GitHub to approve, then come back to this tab."** prompt.
- The **tap** is a fresh gesture (so the new tab is allowed on phones too), copies the code as they leave, and flips the status to "Waiting…".
- The poll runs the whole time, so approval still lands back on its own — nobody has to hunt for the tab.

Also stops claiming *"the code is already filled in"* when GitHub returned no `verification_uri_complete` — it says **"Open GitHub, paste the code, and approve."** in that case, so nobody arrives on GitHub expecting a pre-filled field that isn't there. (I hit exactly this on the #184 preview: the button pointed at plain `github.com/login/device`.)

## Tests

`test/signin-github-tab.test.js` (#179) had **pinned the popup as required**. Flipped it to guard the opposite intent: nothing opens GitHub for the visitor; the anchor tap is the only hop and it's wired to copy + flip status. All 40 offline suites green.

## Verified in a browser
- **On dialog open:** `window.open` calls = **0** (no hijack); code auto-copied; status = "Click Open GitHub to approve…".
- **After tapping Open GitHub:** still no scripted open (native `target=_blank`); status → "Waiting for you to approve…"; href carries the pre-filled code.

### DIFF
--- a/server/signin.js
+++ b/server/signin.js
@@ -152,15 +152,22 @@
         }
 
         var uri = d.verification_uri_complete || d.verification_uri;
+        var prefilled = !!d.verification_uri_complete;
         var where = document.getElementById('tds-where');
-        // A native target=_blank anchor is the only hop that every browser
-        // honours: a scripted window.open() fired after this await is outside
-        // the click's activation window, so Safari and in-app webviews either
-        // swallow it or, worse, open it in this tab. Losing this tab loses the
-        // poll below, and the sign-in can never complete. The anchor cannot
-        // navigate the page it sits on.
+        // One hop to GitHub, and the visitor takes it. We deliberately do not
+        // open it for them: a dialog that yanks you to a fresh GitHub tab
+        // the instant it appears reads as a hijack, not a sign-in — and on
+        // desktop that scripted open actually lands, so the visitor is gone
+        // before they've read a word. A native target=_blank anchor is the only
+        // hop every browser honours anyway (a scripted open() after this await
+        // is outside the click's activation window, so Safari and in-app
+        // webviews swallow it or open it in this tab, losing the poll). The tap is a
+        // fresh gesture, so the new tab is allowed on phones too, and it copies
+        // the code as the visitor leaves. The anchor cannot navigate this page.
         if (isGithubHttpsUrl(uri)) {
-          where.innerHTML = 'Open GitHub and approve. The code is already filled in.'
+          where.innerHTML = (prefilled
+              ? 'Open GitHub and approve — the code is already filled in.'
+              : 'Open GitHub, paste the code, and approve.')
             + '<a class="tds-open" id="tds-open" href="' + esc(uri) + '"'
             + ' target="_blank" rel="noopener noreferrer">'
             + '<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M12 .297c-6.63 0-12 5.373-12 12 0'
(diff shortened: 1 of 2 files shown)


serenakeyitan (approved):
<h3>Breeze Summary</h3>

Follow-up to #179 / #183 / #184: the sign-in dialog no longer auto-`window.open()`s GitHub. The visitor taps a native `target=_blank` **Open GitHub** anchor; that tap copies the device code and flips status to waiting; the poll still starts immediately. Copy is honest when GitHub omitted `verification_uri_complete`.

- `window.open` is gone from `server/signin.js`; the only hop is an `isGithubHttpsUrl`-gated `<a target="_blank" rel="noopener noreferrer">`.
- Click handler copies if needed and does not `preventDefault`, so the browser opens the tab.
- Poll still starts whether or not the visitor has tapped.
- Tests now forbid the popup instead of requiring it. Local `signin-github-tab`, `overlay-pure`, `browser-bundles-parse`, and `tdoc-start` pass; CI offline suite + CodeQL are green.
- Worker still forwards `verification_uri_complete`; the client uses `!!d.verification_uri_complete` for the paste vs pre-filled sentence.
- Overlay / onboard / landing still share `__tdocSignIn` — no second device-flow client.

Approve. Not 5 because the dialog's first paint still says "Opening GitHub…" during `device/start`, and the paste-vs-prefilled copy is not pinned in tests.

| Filename | Overview |
|---|---|
| `server/signin.js` | Drop auto-open; visitor-driven anchor; honest prefilled copy |
| `test/signin-github-tab.test.js` | Guard the opposite of #179: nothing opens GitHub for the visitor |

<h3>Issues</h3>

1. **`server/signin.js` — leftover "Opening GitHub…" placeholder (nit)**  
   `#tds-where` still paints as "Opening GitHub…" until `/api/auth/device/start` returns (often 1–2s). That is the same hijack language this PR removes, and it is the first thing the visitor sees. Swap it for "Starting…" or similar so the first frame matches the new intent. Same file still comments `isGithubHttpsUrl` as "Only ever auto-open github.com".

2. **`test/signin-github-tab.test.js` — paste-vs-prefilled copy unpinned (nit)**  
   The second stated fix (do not claim the code is filled in when `verification_uri_complete` is missing) has no assert. This file is already source-grep; one line on `paste the code` / `prefilled` would lock it. Leftover error strings still talk about "the popup".

No blocking issues.

---

yayashuxue:
🌐 **Live preview:** https://pr-186-tdoc-preview.jyshi1107.workers.dev — Sign in with GitHub: the dialog no longer auto-jumps to GitHub; it waits for the visitor to tap **Open GitHub**, with the code already on the clipboard.

Issue #1, opened 2026-06-05
[REVIEW] reconcileAnchors (publish-time anchor rebinding) has zero tests despite being the most failure-prone correctness path


serenakeyitan:
**Severity:** P1  |  **Dimension:** coverage-critical-path  |  **Confidence:** 9/10
**Location:** `/Users/keyitan/.claude/skills/tdoc/worker/worker.js:251-317`
**Confirmed by:** testing (survived adversarial verification)

### Problem
reconcileAnchors is the heuristic that, at publish time, re-binds every comment's element anchor against the NEW version's artifact set (worker.js:251). It was shipped specifically to fix the anchor-loss bug in commit 8ab5b15 ('preserve aid history across versions so applied comments stay anchored'). Its logic is genuinely subtle: fingerprint+heading candidate filtering (273-276), single-candidate adoption (278), tag-only fallback when zero candidates (280-282), and a 'lost' marker with ambiguous/no_candidate reasons (303-313). Every branch is a place where a comment can silently jump to the wrong artifact or be marked lost. None of it is exercised by any test — comment-history.test.js loads snapshotAt/historyList but never reconcileAnchors. A regression here (e.g. adopting the wrong candidate, or losing an anchor that should rebind) ships silently because the only validation is manual eyeballing of a live doc.

### Systematic fix (not a patch)
Add a pure unit test that slices reconcileAnchors + its deps out of worker.js (same VM-sandbox technique comment-history.test.js already uses) and drives it with synthetic comment lists + aidsInVersion arrays. Cover: (a) anchor already valid in new version → no event appended; (b) exactly one fingerprint+heading candidate → anchor_changed to that aid; (c) zero candidates but one tag-only match → tag-only adoption; (d) >1 candidate → marked lost with reason 'ambiguous'; (e) zero candidates and zero tag matches → lost with reason 'no_candidate'; (f) older-version folds still resolve after a rebind event is appended (the actual cross-version guarantee from 8ab5b15). Assert events are append-only and at_version == publish version.

---
From the 2026-06 engineering review (`docs/ENGINEERING-REVIEW-2026-06.md`). Finding id: `reconcile-anchors-untested`.

serenakeyitan:
Fixed in v0.4.0 (https://github.com/serenakeyitan/tdoc/releases/tag/v0.4.0). See the 2026-06 review fixes on main.

PR #188, opened 2026-08-20
fix: park the comment popup at the selection end, not the line origin

yayashuxue:
## Problem

On a multi-line text selection the new-comment sheet sat at the **left of the wrap**, not where the mouse/caret finished. `Range.getBoundingClientRect()` is the union of every line box.

## Fix

- Pick the live line box under mouse-up / caret X (`getClientRects`), not the union.
- Park the sheet at that X (selection end), clamped in document coordinates so a caret near the right edge still keeps the 320px sheet on screen.
- Pins / fallback Y use the first visible line box so a WebKit 0×0 union does not send them to the top of the page.

## Tests

- `overlay-pure`: nearest-line math + end-rect X is the caret, not `line.left`.
- `ui.test.js`: multi-rect `<h2>`+`<p>` selection; popup is on the last line and nearer the caret than the line origin.

Review of the local diff: 0 bugs, 4 non-blocking suggestions (pointer X clamp on a short last line, stronger Playwright Y check, shorter comments, `commentY` empty-rect fallback). Shipping as-is.

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -120,6 +120,55 @@
     box.style.height = (r.height + inset * 2) + 'px';
   }
 
+  // Range.getBoundingClientRect() is the UNION of every line box. A wrapped
+  // selection is therefore a tall rectangle whose bottom-left is nowhere near
+  // the caret; WebKit also sometimes returns a 0×0 rect when the range
+  // crosses a block. The new-comment popup and pin Y used that union, so the
+  // card "sometimes" sat off the highlight. Prefer the live line box.
+  function isVisibleClientRect(r) {
+    return !!r && (r.width > 0 || r.height > 0);
+  }
+  function nearestClientRect(rects, x, y) {
+    let best = null, bestD = Infinity;
+    const list = rects && rects.length != null ? rects : [];
+    for (let i = 0; i < list.length; i++) {
+      const r = list[i];
+      if (!isVisibleClientRect(r)) continue;
+      const dx = x - Math.max(r.left, Math.min(x, r.right));
+      const dy = y - Math.max(r.top, Math.min(y, r.bottom));
+      const d = dx * dx + dy * dy;
+      if (d < bestD) { bestD = d; best = r; }
+    }
+    return best;
+  }
+  function firstVisibleClientRect(target) {
+    if (!target) return null;
+    if (target.getClientRects) {
+      const rects = target.getClientRects();
+      for (let i = 0; i < rects.length; i++) {
+        if (isVisibleClientRect(rects[i])) return rects[i];
+      }
+    }
+    if (!target.getBoundingClientRect) return null;
+    const r = target.getBoundingClientRect();
+    return isVisibleClientRect(r) ? r : null;
+  }
+  function clientRectNearPoint(target, x, y) {
+    if (!target) return null;
+    if (target.getClientRects && Number.isFinite(x) && Number.isFinite(y)) {
+      const near = nearestClientRect(target.getClientRects(), x, y);
+      if (near) return near;
+    }
+    return firstVisibleClientRect(target);
+  }
+  // A zero-width box on `line` at x — the caret / mouse-up, not the line origin.
+  function endRectOnLine(line, x) {
+    if (!line) return null;
+    const left = Number.isFinite(x) ? x : line.right;
+    const height = line.height || (line.bottom - line.top) || 0;
+    return { left, right: left, top: line.top, bottom: line.bottom, width: 0, height };
+  }
+
   // ========== Styles ==========
   // Each logical group is one comment block; rules within a group are tightly
   // packed. The narrow visual mode lives at the bottom and overrides base.
@@ -2237,7 +2286,7 @@
     const mark = state.anchorMarks.get(c.id);
     if (mark && (mark.ranges?.[0] || mark.el)) {
       const target = mark.ranges?.[0] || mark.el;
-      const r = target.getBoundingClientRect();
+      const r = firstVisibleClientRect(target) || target.getBoundingClientRect();
       // For element anchors expose the element + its rect so renderPins can
       // spread multiple comments DOWN a tall element instead of stacking them
       // all at its top edge. (targetEl is the live element; el may be the outline.)
@@ -2592,9 +2641,9 @@
     let anchorRect = null;
     // Prefer the underlying TARGET ELEMENT (canvas/img/video etc) over the
     // overlay outline div — same rect, but more semantically correct.
-    if (mark.ranges?.[0]) anchorRect = mark.ranges[0].getBoundingClientRect();
-    else if (mark.targetEl?.getBoundingClientRect) anchorRect = mark.targetEl.getBoundingClientRect();
-    else if (mark.el?.getBoundingClientRect) anchorRect = mark.el.getBoundingClientRect();
+    if (mark.ranges?.[0]) anchorRect = firstVisibleClientRect(mark.ranges[0]);
+    else if (mark.targetEl) anchorRect = firstVisibleClientRect(mark.targetEl);
+    else if (mark.el) anchorRect = firstVisibleClientRect(mark.el);
     if (!anchorRect) return;
 
     // We consider the anchor "comfortably visible" if its top is between the
@@ -3273,13 +3322,18 @@
     // body. The caller signals this by setting anchor._placeAbove = true.
     document.body.appendChild(popup);   // append first so offsetHeight is known
     const popupH = popup.offsetHeight || 140;
+    const popupW = popup.offsetWidth || 320;
     if (anchor._placeAbove && rect.top - 8 - popupH >= 8) {
       popup.style.top = (window.scrollY + rect.top - popupH - 8) + 'px';
     } else {
       popup.style.top = (window.scrollY + rect.bottom + 8) + 'px';
     }
-    const left = Math.min(rect.left + window.scrollX, window.innerWidth - 340);
-    popup.style.left = Math.max(8, left) + 'px';
+    // `rect.left` is the caret / mouse-up X for text selections (not the
+    // line-box origin). Clamp in document coords so a caret near the right
+    // edge still keeps the 320px sheet on screen.
+    const maxLeft = window.scrollX + window.innerWidth - popupW - 8;
+    const left = Math.min(Math.max(window.scrollX + 8, window.scrollX + rect.left), maxLeft);
+    popup.style.left = left + 'px';
 
     if (anchor.kind === 'text' && anchor._range) {
       setPendingTextHighlight(anchor._range);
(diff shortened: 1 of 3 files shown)


server/overlay.js
```diff
@@ -120,6 +120,55 @@
...
+  }
+  function nearestClientRect(rects, x, y) {
+    let best = null, bestD = Infinity;
+    const list = rects && rects.length != null ? rects : [];
+    for (let i = 0; i < list.length; i++) {
+      const r = list[i];
+      if (!isVisibleClientRect(r)) continue;
+      const dx = x - Math.max(r.left, Math.min(x, r.right));
+      const dy = y - Math.max(r.top, Math.min(y, r.bottom));
+      const d = dx * dx + dy * dy;
```
serenakeyitan, line 139: Nit: Euclidean distance to the clamped point on the box picks the previous full-width line when mouse-up is in the empty space to the right of a short last line (typical wrap-width drag).

With line1 `[40,400]×[100,120]` and line2 `[40,180]×[120,140]`, pointer `(200, 130)` — 20px past the last character — snaps to line1. Prefer the line band that contains `y` (min `dy`), then `dx`.

`ui.test.js` uses `last.right - 2`, so it does not catch this. Fine to land as-is.

test/ui.test.js
```diff
@@ -352,6 +352,61 @@ async function tPub(name, fn) {
...
+      }));
+    }, { x: pos.x, y: pos.y });
+    await page.waitForSelector('.tdoc-popup', { timeout: 2000 });
+    const popup = await page.$eval('.tdoc-popup', el => {
+      const r = el.getBoundingClientRect();
+      return { top: r.top, left: r.left };
+    });
+    // Opened below the caret line (last.bottom + 8). Must not sit on the first
+    // line of a multi-block selection.
+    if (popup.top < pos.lastTop - 4) {
```
serenakeyitan, line 398: Nit: `popup.top < lastTop - 4` still passes for the old union-bottom placement, because union.bottom is the last line's bottom on a normal multi-line range. The X check is the real gate here, and it only covers caret-on-glyph (`last.right - 2`), not wrap-width overhang.

serenakeyitan (approved):
<h3>Breeze Summary</h3>
Parks the new-comment sheet on the live line box under mouse-up / caret, instead of `Range.getBoundingClientRect()`'s union (which put a wrapped selection's popup at the wrap's left). Pin / fallback Y uses the first visible line box so a WebKit 0×0 union does not send cards to the top of the page.

- `overlay-pure` helpers (`isVisibleClientRect` / `nearestClientRect` / `endRectOnLine`) pass locally (24/24); CI `offline test suite` is green.
- `openPopup` now clamps in document coordinates (`scrollX + innerWidth - popupW`) instead of mixing viewport `innerWidth - 340` with document `left`.
- `body.tdoc-narrow .tdoc-popup { left: 12px !important }` still wins on phones, so the new X math does not fight the drawer sheet.
- Overlay remains the single source of truth (`server/overlay.js` inlined into the worker); no second copy to update.
- Playwright covers caret-on-last-line vs line origin. I did not run `ui.test.js` here (playwright not installed in this workspace).

Approve. Net user-visible fix, helper tests exist, offline CI is green. Not 5 because the gated Playwright suite was not run here, and wrap-width mouse-up on a short last line still nearest-picks the previous full-width line (see #1).

| Filename | Overview |
| --- | --- |
| `server/overlay.js` | Line-box helpers + `selectionEndRect`; popup X at caret; pin Y uses first visible rect |
| `test/overlay-pure.test.js` | Unit tests for visible / nearest / end-rect math |
| `test/ui.test.js` | Playwright: multi-rect `<h2>`+`<p>` selection parks nearer the caret than the line origin |

<h3>Issues</h3>

No blocking issues.

1. **`server/overlay.js`** — Nit: Euclidean nearest-rect loses the short last line on wrap-width mouse-up.
   With line1 `[40,400]×[100,120]` and line2 `[40,180]×[120,140]`, pointer `(200, 130)` (20px past the last character) snaps to line1. A wrap-width drag-select commonly releases in that overhang. Prefer the line band that contains `y` (min `dy`), then `dx`. Fine to land as-is; a follow-up would lock the gesture the PR is named for.

2. **`test/ui.test.js`** — Nit: the Y check (`popup.top < lastTop - 4`) does not distinguish last-line from union-bottom (they match on a normal multi-line range), and X is `last.right - 2`, so the overhang in #1 cannot fail this test. Assert `popup.top >= last.bottom` (below the caret line) and add an overhang mouse-up if you fix #1.

3. **`CHANGELOG.md`** — Nit: this is a user-visible overlay fix and Unreleased has no line for it.

---

server/overlay.js
```diff
@@ -120,6 +120,55 @@
...
+  function isVisibleClientRect(r) {
+    return !!r && (r.width > 0 || r.height > 0);
+  }
+  function nearestClientRect(rects, x, y) {
+    let best = null, bestD = Infinity;
+    const list = rects && rects.length != null ? rects : [];
+    for (let i = 0; i < list.length; i++) {
+      const r = list[i];
+      if (!isVisibleClientRect(r)) continue;
+      const dx = x - Math.max(r.left, Math.min(x, r.right));
```
yayashuxue, line 137: **[bug]** `nearestClientRect` uses Euclidean distance to the box. On a wrapped selection the last line is usually shorter than the one above it. A mouse-up on the last line’s Y but to the right of `last.right` (the common wrap-drag: pull down the right of the column) is inside the previous line’s X-span and only ~one line-height below it, so distance to the long previous line is ~line-height while distance to the short last line is hundreds of pixels. `selectionEndRect` prefers the pointer path and never reaches the caret fallback, so the sheet’s `rect.bottom` is the previous line’s bottom — not the line the user released on. The Playwright test cannot catch this: it fires mouseup at `last.right - 2`, which is inside the last box.

**Suggestion:** Pick the line by vertical overlap first (rects whose `[top, bottom]` contain `pointerY`, then nearest X among those), and only then fall back to Euclidean. Keep unclamped `pointerX` so a wrap-release to the right of a short last line still parks at mouse-up X, on the last line’s Y. Add a case at last-line Y and previous-line right X.

server/overlay.js
```diff
@@ -2235,7 +2284,7 @@
     const mark = state.anchorMarks.get(c.id);
     if (mark && (mark.ranges?.[0] || mark.el)) {
       const target = mark.ranges?.[0] || mark.el;
-      const r = target.getBoundingClientRect();
+      const r = firstVisibleClientRect(target) || target.getBoundingClientRect();
```
yayashuxue, line 2289: **[suggestion]** `firstVisibleClientRect` already returns null when both `getClientRects()` and `getBoundingClientRect()` are empty/0×0. The `|| target.getBoundingClientRect()` fallback then reintroduces that 0×0 union, so `commentY` still writes `y = scrollY + 0` and a pin can jump to the top of the page — the WebKit hole this helper was meant to close — whenever *all* rects are empty.

**Suggestion:** If `firstVisibleClientRect` is null, skip the live rect and use `c.anchor.fallback` (the branch already below) instead of stuffing a 0×0 bounding rect back in.

server/overlay.js
```diff
@@ -3350,8 +3404,8 @@
     const articleTop = articleEl.getBoundingClientRect().top + window.scrollY;
     const articleHeight = Math.max(1, articleEl.scrollHeight);
     let rect = null;
-    if (anchor.kind === 'text' && anchor._range) rect = anchor._range.getBoundingClientRect();
-    else if (anchor.kind === 'element' && anchor._el) rect = anchor._el.getBoundingClientRect();
+    if (anchor.kind === 'text' && anchor._range) rect = firstVisibleClientRect(anchor._range);
```
yayashuxue, line 3409: **[suggestion]** `captureFallbackPosition` now takes the first visible line box and uses its vertical midpoint. For a healthy multi-line range that is no longer the union’s center — a 10-line highlight’s saved `fallback.ratio` slides up to line 1. The 0×0 WebKit case is worth special-casing; a visible union’s center was the better estimate of “where this comment was” if the text later fails to re-anchor.

**Suggestion:** Use `range.getBoundingClientRect()` when `isVisibleClientRect` accepts it; use `firstVisibleClientRect` only when the union is empty.

server/overlay.js
```diff
@@ -120,6 +120,55 @@
     box.style.height = (r.height + inset * 2) + 'px';
   }
 
+  // Range.getBoundingClientRect() is the UNION of every line box. A wrapped
```
yayashuxue, line 123: **[suggestion]** The geometry-helper block narrates the bug (union vs caret, WebKit 0×0, “the card sometimes sat off the highlight”). Same for the `openPopup` note that restates `rect.left` is mouse-up X. Comments in this file should stay short WHY, not a changelog of the previous formula.

**Suggestion:** Collapse to one line: prefer `getClientRects()` over the union bounding box because the union’s bottom-left is not the caret, and WebKit may return 0×0 across blocks.

test/ui.test.js
```diff
@@ -352,6 +352,61 @@ async function tPub(name, fn) {
...
+      if (rects.length < 2) return null;
+      const last = rects[rects.length - 1];
+      const sel = window.getSelection();
+      sel.removeAllRanges();
+      sel.addRange(r);
+      return {
+        unionTop: union.top,
+        lastTop: last.top,
+        lastBottom: last.bottom,
+        lineLeft: last.left,
```
yayashuxue, line 380: **[suggestion]** The regression test synthesizes mouseup at `last.right - 2` (inside the last box) and only asserts `popup.top >= lastTop - 4`. That does not fail if the sheet sits under the previous wrap line (`first.bottom + 8` is still below `last.top` for adjacent line boxes), and it never exercises pointer X past `last.right`. `if (!pos) { ... return }` also counts as a pass, so a fixture/live doc without an `<h2>`+`<p>` pair greens the test without checking the bug.

**Suggestion:** Mouseup at `(prev.right - 2, last.top + last.height/2)` and require `popup.top >= last.bottom` (the sheet opens at `rect.bottom + 8`). Treat a missing multi-rect selection as a skip/fail, not `ok()`.

yayashuxue (commented):
## Summary

The change correctly stops using `Range.getBoundingClientRect()` (the union of every line box) for the new-comment sheet: `selectionEndRect` parks on a live `getClientRects()` line at mouse-up / caret X, and `openPopup` now clamps that X in document coordinates. That matches the reported “popup on the left of the wrap” bug when the pointer is actually on the last line’s glyphs. The dominant risk is the line picker itself: Euclidean nearest-rect will snap a wrap-select released in the empty space to the right of a short last line back onto the longer previous line, so Y can land on the wrap above. Pins/fallback now also use the first visible line box — a reasonable WebKit 0×0 guard, but it shifts multi-line fallback `centerY` toward the start of the selection.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 0

Issue #194, opened 2026-08-21
Generated docs have no voice contract, so AI-slop phrasing is the default output


serenakeyitan:
## What problem does this solve?

Nothing in tdoc constrains how the generated prose reads. `SKILL.md` has
extensive rules for the markup (CSP, widget islands, the author-HTML
compatibility contract, responsive defaults) and none for the words inside
it. Whatever voice the model defaults to is what ships.

For most tools that would be a nice-to-have. For tdoc it is a category
weakness: the entire product is "prompt in, document out," so every doc is
model-written, and AI-slop phrasing is the house default unless something
pushes back on it. Binary contrasts, colon reveals, faux-insight setups,
and importance puffery are exactly what a model reaches for when generating
from a short prompt.

There is also no place to put the two adjacent choices that keep coming up
— visual style and section structure — so each one gets re-litigated from
scratch when it surfaces.

## What would you like to happen?

An `authoring/` directory read at generation time, with three slots that
answer three different questions:

- **`voice.md`** — how the prose reads. **Always applied, not selectable.**
  Nobody picks "make it sound like AI," so voice is a floor rather than a
  template a user browses.
- **`style/`** — what the page looks like. Reserved, empty.
- **`structure/`** — which sections a kind of doc has. Reserved, empty.

For voice, vendor the `no-ai-slop` skill by Peter Yang (MIT, ~5.5k stars) —
20+ named patterns, each with a concrete rewrite — rather than writing a
worse list from scratch. Keep the upstream copy verbatim so drift is
detectable, and put tdoc-specific adaptation in a separate file.

Two adaptations are needed, because tdoc differs from the drafts upstream
was written for:

1. **It is an editor; tdoc generates.** Upstream assumes a human pastes a
   draft to be fixed. tdoc has no draft — the constraints have to apply as
   the prose is written, not as a cleanup pass.
2. **"Preserve the writer's voice" has no object.** Much of upstream
   protects a human writer's bluntness, humor, and hedges. A generated doc
   has no such writer. The available anchor is the user's own prompt.

And one hazard worth stating in the file: the rules are written for prose,
but a tdoc doc is full of non-prose — code blocks, identifiers, error
strings, quoted material, table values, SVG labels. A rule that improves a
sentence corrupts an error string. The contract has to fence those off
explicitly.

## Scope

- `authoring/` with the three slots; `style/` and `structure/` ship empty
- `SKILL.md` makes `voice.md` required reading on both generation paths
[...]

PR #195, opened 2026-08-21
Every generated doc goes through a voice contract

serenakeyitan:
Refs #194.

## What

`SKILL.md` has extensive rules for the markup and none for the words inside
it. Since every tdoc doc is model-written, AI-slop phrasing was the house
default with nothing pushing back on it.

Adds `authoring/`, read at generation time, with three slots answering three
different questions:

| Slot | Question | Ships as |
|---|---|---|
| `voice.md` | How does the prose read? | **Always applied** |
| `style/` | What does the page look like? | Empty, reserved |
| `structure/` | Which sections does this kind of doc have? | Empty, reserved |

**Voice is a floor, not a template.** Nobody picks "make it sound like AI,"
so it never appears as something a user browses or selects. `style/` and
`structure/` are mount points only — the layout is settled so adding entries
later is additive, but no entry ships here.

## Why vendor instead of writing our own list

`no-ai-slop` (Peter Yang, MIT, ~5.5k stars) is 20+ named patterns, each with
a concrete rewrite. A from-scratch list would be worse. The copy stays
verbatim under `authoring/vendor/` with its MIT LICENSE, pinned to commit
`d30eddb` **and** to a content sha256, so a local edit and a genuine upstream
refresh are distinguishable. tdoc-specific adaptation goes in `voice.md`,
never by editing the vendored file.

## Two adaptations that mattered

**Upstream is an editor. tdoc generates.** It assumes a human pastes a draft
to be fixed. tdoc has no draft, so the constraints apply as the prose is
written, not as a cleanup pass. `voice.md` also suppresses two upstream
behaviors that belong to editing: asking the reader to paste a draft, and
emitting a "What changed" section.

**"Preserve the writer's voice" has no object.** Much of upstream protects a
human writer's bluntness, humor, and hedges. A generated doc has no such
writer, so that half cannot be followed literally. The available anchor is
the user's own prompt: their nouns, their stance, their level of bluntness.

## The hazard the contract has to fence off

The rules are written for prose, but a tdoc doc is full of non-prose: code
blocks, identifiers, CLI flags, error strings, quoted material, table values,
SVG labels. **A rule that improves a sentence corrupts an error string.**
`voice.md` names those spans as off-limits, and a test asserts it still does.

## Wiring

`voice.md` is referenced from three places, each independently asserted:

- the new top-level **Authoring contract** section in `SKILL.md`
- step 2 of `/tdoc new`
- step 2 of `/tdoc edit` (a regeneration writes new prose, so it applies
  there too — but prose carried over unchanged is left alone)
- the first bullet of **HTML generation rules**

`skills/tdoc/SKILL.md` re-synced, since `agent-md.test.js` asserts the plugin
copy matches. `AGENTS.md` deliberately untouched — it is guarded as an exact
string.

## Tests

`test/authoring.test.js`, 10 assertions, registered in `test/run.js`. It fails
the build if the reference is dropped from either generation path, if the
vendored copy is edited in place, if `voice.md` forks the banned-word list
instead of citing it, or if `style/`/`structure/` gain entries without
`SKILL.md` learning how to select one.

**Negative-verified, not just green:** deleting the `/tdoc new` reference
produced `✗ the reference sits on BOTH generation paths`, and appending one
byte to the vendored file produced `✗ vendored copy was edited in place
(sha 19997d4158f4 vs recorded 16719efd6dc6)`. Both went back to green on
restore. A guard nobody has watched fail is not a guard.

## What this cannot verify

The offline suite has no model, so **no test here proves an agent actually
obeys `voice.md` at generation time.** These tests prove the contract is
wired in and intact — necessary, not sufficient. Behavioral confidence has to
come from generating docs and reading them, which is the natural next step
before this gets leaned on.

## Checklist

- [x] `node test/run.js` — PASS, all 43 suites green
- [x] Negative tests confirm the guards actually fail on breakage
- [x] `skills/tdoc/SKILL.md` re-synced with root
- [x] Upstream MIT LICENSE vendored and credited
- [x] `AGENTS.md` untouched
- [x] No unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ file and `.claude-plugin/plugin.json`.
 
 ### Added
 
+- **Every generated doc goes through a voice contract.** New `authoring/`
+  directory, read at generation time. `authoring/voice.md` applies to
+  `/tdoc new` and every `/tdoc edit` regeneration — a floor, not a
+  user-selectable template, because nobody picks "make it sound like AI."
+  It adapts the vendored `no-ai-slop` rule set (Peter Yang, MIT) to
+  generation rather than editing, names the user's prompt as the voice
+  anchor, and fences off the spans prose rules must never rewrite: code,
+  identifiers, quoted material, and data. `authoring/style/` and
+  `authoring/structure/` ship as empty reserved mount points. `#194`.
 - **PR preview Worker (`tdoc-preview`).** Pull requests on tornado-doc/tdoc
   get a unique `pr-<N>` preview URL and a sticky comment pointing at
   `/d/conway-life/v/2` on that host — published chrome, not Local Studio,
--- a/SKILL.md
+++ b/SKILL.md
@@ -223,16 +223,19 @@ sleep 1
 
 ## Authoring contract — read before writing any doc
 
-**`authoring/voice.md` is required reading before you write doc HTML**, on
-every `/tdoc new` and every regeneration in `/tdoc edit`. It is a floor, not
+**`$SKILL_DIR/authoring/voice.md` is required reading before you write doc
+HTML**, on every `/tdoc new` and every regeneration in `/tdoc edit`.
+`$SKILL_DIR` is the installed skill directory resolved in "Setup check"
+above (`~/.claude/skills/tdoc`, or `~/.codex/skills/tdoc` under Codex) —
+**not** the current working directory, which is the user's project. It is a floor, not
 a user-selectable option: there is no switch, and no doc is exempt.
 
 It carries tdoc's adaptation of the vendored `no-ai-slop` rule set
-(`authoring/vendor/no-ai-slop.md`) — which prose the rules govern, which
+(`$SKILL_DIR/authoring/vendor/no-ai-slop.md`) — which prose the rules govern, which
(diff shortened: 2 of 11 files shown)


serenakeyitan:
Follow-up commit `cc14f85` fixes a defect in the first pass that would have made this whole PR inert.

The references were written as bare relative paths (`authoring/voice.md`). The agent's working directory is the **user's project**, not the skill install, so that path resolves to nothing. The agent would read no contract, generation would look completely normal, and nothing would surface the failure. Silent, and invisible in review.

The original tests did not catch it: they asserted the string was *present*, not that it was *reachable*. Two guards added:

- every real `authoring/` path in `SKILL.md` must be `$SKILL_DIR`-anchored, reporting the offending line number
- the Authoring contract section must state where `$SKILL_DIR` points and warn that it is not the cwd, so the variable is not left dangling

Negative-verified: reverting one reference to a bare path fails with `bare authoring/ paths in SKILL.md: authoring/voice.md (line 247)`.

Worth noting for anyone reviewing the test file: this is the class of bug these tests exist for. A contract nothing can reach is indistinguishable from no contract at all, and only a reachability assertion tells them apart.

Issue #197, opened 2026-08-21
authoring/style/ has no default, so link/danger/success colors are improvised per doc


serenakeyitan:
## What problem does this solve?

`authoring/style/` shipped in #195 as an empty mount point. Empty means
every doc falls through to the overlay's default reading template, and the
parts that template does not cover — link color, success color, destructive
color — are improvised per doc.

tdoc already has a design system for this. It is documented at
`tdoc.dev/d/tdoc-design-tokens/v/1`, which states outright: *"The
hand-built minimal system behind tdoc — no external template. This page
uses the tokens it documents."* It is in production for the product chrome
but nothing carries it into generated docs.

## What would you like to happen?

Extract it as `authoring/style/default.md`, applied when a doc selects no
style — which today is every doc.

**Color (9 tokens, authoritative — the doc's prose and its `:root` agree):**

| Token | Value | Use |
|---|---|---|
| `--td-accent` | `#1652f0` | links + the one primary button |
| `--td-accent-hover` | `#1245d0` | primary hover |
| `--td-ink` | `#111` | primary text |
| `--td-muted` | `#666` | secondary / meta |
| `--td-line` | `#eee` | thin borders / dividers |
| `--td-surface` | `#f7f7f7` | subtle fills |
| `--td-danger` | `#b42318` | destructive only |
| `--td-danger-tint` | `#fdeceb` | danger hover-fill |
| `--td-ok` | `#087443` | success / confirmed |

**Typography:** `max-width: 720px`, body `15px/1.6` system-ui, h1
`29px/700/-.01em`, meta `12px`, mono `ui-monospace, "SF Mono", Menlo`.

**Principles:** restraint (the Notion / Linear / Google Docs lineage — when
in doubt, remove); ration the accent (blue only for links and the single
primary action); quiet the dangerous (destructive actions hide behind a `⋯`,
red only on reveal); chrome never fights content; no web fonts, no framework.

## The typography change is deliberate and worth stating plainly

The overlay's default reading template is body **17px**/1.6 with h1 **34px**.
These tokens are body **15px** with h1 **29px**. Adopting them wholesale
makes every newly generated doc denser than docs generated before it.

That is the intended outcome: tdoc's documents and tdoc's product chrome
become one system rather than two. Recording it here because the diff alone
looks like a small number change, and it is not — it is a reading-experience
decision that applies to every future doc.

Docs already published are unaffected; their HTML is a build artifact that
was written with the old defaults.

[...]

PR #111, opened 2026-08-14
chore: auto-review every human PR (including updates)

serenakeyitan:
## Why

`tornado-doc/tdoc` already has `greptile.json`, but Greptile stopped posting after the repo moved to the org — no GitHub App is installed on `tornado-doc` (last review was PR #83 on 2026-07-17). Collaborator PRs from @yayashuxue have been landing without an automatic audit comment.

This PR only updates the repo config so that **once Greptile Apps is reinstalled on `tornado-doc` → `tdoc`**, every human PR (open + each new push, including drafts) gets a review comment. Dependabot/bots are excluded.

## Config

- `triggerOnUpdates: true` — re-review on every push
- `triggerOnDrafts: true` — do not wait for ready-for-review
- `shouldUpdateDescription: false` — keep the existing “comment, don’t rewrite the PR body” behavior
- `commentTypes: logic + syntax` — audit, not style nits
- `excludeAuthors: dependabot / *-bot`

## Follow-up (needs a one-click org admin install)

Install Greptile on the org (only this repo):
https://github.com/apps/greptile-apps/installations/new/permissions?target_id=309246106

Then Greptile dashboard → enable `tornado-doc/tdoc`. After that I will retrigger review on the currently open PRs (#109, #99).

PR #198, opened 2026-08-21
Docs carry tdoc's own design system by default

serenakeyitan:
Refs #197. Follows #195, which shipped `authoring/style/` as an empty mount point.

## What

Extracts tdoc's own design system — documented and already in production at
`tdoc.dev/d/tdoc-design-tokens/v/1` — into `authoring/style/default.md`,
applied when a doc selects no style. Today that is every doc.

That page states its own status plainly: *"The hand-built minimal system
behind tdoc — no external template. This page uses the tokens it
documents."* So this is extraction, not invention.

**Nine color tokens, naming roles rather than colors:**

| Token | Value | Use |
|---|---|---|
| `--td-accent` | `#1652f0` | links + the one primary button |
| `--td-accent-hover` | `#1245d0` | primary hover |
| `--td-ink` | `#111` | primary text |
| `--td-muted` | `#666` | secondary / meta |
| `--td-line` | `#eee` | thin borders / dividers |
| `--td-surface` | `#f7f7f7` | subtle fills |
| `--td-danger` | `#b42318` | destructive only |
| `--td-danger-tint` | `#fdeceb` | danger hover-fill |
| `--td-ok` | `#087443` | success / confirmed |

Plus a system-font type scale and the five principles that decide what the
tokens do not: restraint, ration the accent, quiet the dangerous, chrome
never fights content, no framework.

## The density change is the part to look at

| | Before (overlay default) | After (house style) |
|---|---|---|
| body | 17px / 1.6 | **15px** / 1.6 |
| h1 | 34px | **29px** |

Newly generated docs get denser. That is deliberate — it is the scale the
product chrome already uses, and one system beats two. Flagging it because
the diff looks like a small number change and it is not: it applies to every
future doc.

**Docs already published are unaffected.** Their HTML is a build artifact
written under the old defaults.

## A contradiction this forced

`SKILL.md` said **"Don't write your own CSS for these"** while the house
style *is* a CSS block every doc carries. Both could not stand.

The rule is now: apply the house style's tokens, invent no scale or palette
of your own, and scope any doc-specific CSS tightly. A test asserts the old
blanket rule cannot come back.

## Deliberately not touched

Nothing in `server/overlay.js` or the worker. The overlay's injected
defaults remain the floor a doc falls to for anything the house style does
not set; this only changes what the agent *writes into* the doc. That keeps
a hot file — currently edited across several open PRs and one other working
session — entirely out of this change.

## Tests

Four new guards in `test/authoring.test.js`, each negative-verified by
breaking it and watching it fail:

| Guard | Broken by | Fails with |
|---|---|---|
| Every `style/` entry is selectable from `SKILL.md` | adding `editorial.md` unreferenced | `style entries SKILL.md never mentions: editorial.md` |
| `default.md` still defines its tokens | emptying the file | `default.md no longer defines --td-accent` |
| The default is read on the generation path | — | `/tdoc new does not read authoring/style/default.md` |
| The old CSS ban stays gone | re-adding the sentence | `the old blanket "Don't write your own CSS" rule is back` |

The first is the one that matters longer-term: it is the single-source-of-truth
guard. A style entry the agent cannot be told to use, or a name a user picks
that the agent never heard of, both fail **silently** — the doc still
generates, just with the wrong look. Same failure shape as the bare-path bug
in #195.

## Checklist

- [x] `node test/run.js` — PASS, all 43 suites green
- [x] All four new guards negative-verified
- [x] `skills/tdoc/SKILL.md` re-synced with root
- [x] `AGENTS.md` untouched
- [x] `server/overlay.js` and `worker/` untouched
- [x] No unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ file and `.claude-plugin/plugin.json`.
 
 ### Added
 
+- **Docs now carry tdoc's own design system.** New
+  `authoring/style/default.md`, extracted from
+  `tdoc.dev/d/tdoc-design-tokens/v/1` and applied when a doc selects no
+  style -- which today is every doc. Nine `--td-*` color tokens (accent
+  reserved for links and the single primary action; red reserved for
+  destruction), a system-font type scale, and the principles that decide
+  what the tokens do not. **Newly generated docs are denser than before**
+  (body 15px rather than the overlay's 17px, h1 29px rather than 34px):
+  deliberate, so documents and product chrome are one system. Docs already
+  published are unaffected. `SKILL.md`'s blanket "don't write your own CSS"
+  rule is replaced -- the house style *is* CSS every doc carries; what
+  stays forbidden is inventing a scale or palette on top of it. `#197`.
 - **Every generated doc goes through a voice contract.** New `authoring/`
   directory, read at generation time. `authoring/voice.md` applies to
   `/tdoc new` and every `/tdoc edit` regeneration — a floor, not a
--- a/SKILL.md
+++ b/SKILL.md
@@ -223,29 +223,43 @@ sleep 1
 
 ## Authoring contract — read before writing any doc
 
-**`$SKILL_DIR/authoring/voice.md` is required reading before you write doc
-HTML**, on every `/tdoc new` and every regeneration in `/tdoc edit`.
+Two files are required reading before you write doc HTML, on every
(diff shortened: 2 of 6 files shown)


serenakeyitan:
Closing — the extracted 15px design-token style isn't the right default. Replacing it with a style extracted from tdoc.dev/d/avgraph-auto-research/v/1. Reopening under a fresh branch.

PR #231, opened 2026-08-23
bar: make ⋯ the permanent home for Duplicate / Download

### LINKED ISSUE #230: Collapse Copy / Duplicate / Download into the ⋯ menu at every width, not just <900px
The bar already has the ⋯ menu this needs. It is gated behind a breakpoint, and Copy never enters it at any width.

## Current behavior

On a published doc at ≥900px, the right side of the bar carries three secondary controls, two of which open their own dropdown:

```
theme · Copy ▾ (Doc only / Doc + comments) · Duplicate · Download ▾ (HTML / PDF) · Share · avatar
```

Three menus competing in one bar. Duplicate and Download are rare actions next to Share, but they take permanent inline space.

## What already exists

`server/overlay.js:1020` builds the ⋯ button (`#tdoc-more-btn`) and `#tdoc-secondary-menu`, already populated with `duplicate`, `download`, `download-pdf` (and `saveas` on forks), plus the version group.

It is hidden by default:

```css
/* overlay.js:475 */
.tdoc-bar .tdoc-secondary-toggle { display: none; padding: 6px 10px; }
```

and revealed only by two rules, which also hide the inline buttons:

```css
/* overlay.js:786-787, inside @media (max-width: 900px) */
.tdoc-bar #tdoc-duplicate-btn, .tdoc-bar #tdoc-download-wrap, .tdoc-bar #tdoc-saveas-btn { display: none; }
.tdoc-bar .tdoc-secondary-toggle { display: inline-flex; }

/* overlay.js:799-800 */
body.tdoc-narrow .tdoc-bar #tdoc-duplicate-btn, … { display: none; }
body.tdoc-narrow .tdoc-bar .tdoc-secondary-toggle { display: inline-flex; }
```

So below 900px the bar already looks the way this issue asks for. Above 900px it does not.

## Copy is the odd one out

`#tdoc-copy-md-btn` is never in the ⋯ menu at any width. At <700px only its label span hides:

```css
/* overlay.js:793 */
.tdoc-bar #tdoc-copy-md-btn span { display: none; }
```

leaving an icon-only Copy dropdown sitting next to ⋯ on phones — two separate menus on the narrowest screen, which is where there is least room for them.

## Proposal

1. Show ⋯ at every width: drop the `display:none` default (overlay.js:475) and the two overrides that reveal it.
2. Hide inline `#tdoc-duplicate-btn` / `#tdoc-download-wrap` / `#tdoc-saveas-btn` unconditionally. The menu items already exist, so this is mostly deleting CSS.
3. Fold Copy's two modes into ⋯ as `Copy as Markdown` and `Copy with comments`, and drop the standalone control. This is the one part that is not just CSS — `#tdoc-copy-md-menu`'s `data-mode` handler needs to be reachable from `#tdoc-secondary-menu`.

Result, at every width:

```
theme · Share · ⋯ · avatar
```

with ⋯ holding: Version (when >1) · Copy as Markdown · Copy with comments · Duplicate · Download HTML · Download PDF.

## Open question

The version group inside ⋯ is currently hidden above 700px because the inline version chip covers it:

```css
/* overlay.js:466-467 */
.tdoc-secondary-menu .tdoc-sec-versions { display: none; }
@media (max-width: 700px) { .tdoc-secondary-menu .tdoc-sec-versions { display: block; } }
```

If ⋯ becomes permanent, that rule should be a deliberate choice rather than an inherited one. Keeping the inline chip *and* the menu group seems right — switching versions is a reading action, not a management action, so it earns its inline spot in a way Duplicate and Download do not.

unknowncici:
Closes #230 (partially — see *Not in this PR* below).

## What & why

The ⋯ overflow menu already carried `duplicate` / `download` / `download-pdf` (and `saveas` on forks), but `.tdoc-secondary-toggle` was `display: none` by default and only revealed at `max-width: 900px` or under `body.tdoc-narrow`. Above 900px the bar instead carried Duplicate and a Download dropdown inline:

```
before (≥900px):  theme · Copy ▾ · Duplicate · Download ▾ · Share · avatar
after  (all):     theme · Copy ▾ · Share · ⋯ · avatar
```

Three secondary controls and two competing menus sat next to Share on desktop, while the phone layout already looked the way we'd want. This makes the phone behavior the behavior everywhere.

## The change

`server/overlay.js`, CSS only:

- `.tdoc-bar .tdoc-secondary-toggle` → `display: inline-flex` (was `none`)
- `.tdoc-bar #tdoc-duplicate-btn, #tdoc-download-wrap, #tdoc-saveas-btn` → `display: none`, unconditionally
- the two reveals that are now redundant are removed: the pair inside `@media (max-width: 900px)`, and the `body.tdoc-narrow` pair
- the breakpoint-ladder comment gains a line saying ⋯ is unconditional

Net −6 / +8 lines, no JS touched.

The inline buttons stay in the markup rather than being deleted. `test/duplicate-download.test.js` asserts the overlay source contains `id="tdoc-duplicate-btn"` / `id="tdoc-download-btn"`, `ui.test.js` asserts they exist with the right labels on published chrome, and `hosted-oob-behavior.test.js` asserts they are *absent* from an exported file. Hiding rather than removing keeps all three contracts intact.

## Tests

`npm test` — 43/43 suites green, which is what `test.yml` runs on `pull_request`.

The browser suites could not be run locally: `npx playwright install chromium` fails with `Playwright does not support chromium on mac13-arm64`. Reviewing what they would have covered:

- `responsive.test.js` asserts ⋯ **is** visible in narrow mode (still true) and never asserts it is hidden in wide mode, so the change is compatible.
- `ui.test.js` reaches Duplicate/Download with `page.$` and reads `textContent` — presence and labels only, no visibility check and no click — so hiding them keeps it green.
- Both suites' published-chrome assertions are `tPub`, which skips against the local fixture anyway.

If a maintainer with a working Playwright install can run `npm run test:all`, that would be worth confirming rather than taking my reading of it.

## Not in this PR

Step 3 of #230 — folding Copy's two modes into ⋯ and dropping the standalone control. It needs `#tdoc-copy-md-menu`'s `data-mode` handler reachable from `#tdoc-secondary-menu`, and it changes `ui.test.js`, which drives `#tdoc-copy-md-btn` with `page.click` and therefore needs it visible. That is a JS + test change and reviews better on its own.

Also left alone deliberately: the version group inside ⋯ stays hidden above 700px (`overlay.js:466-467`), so the inline version chip remains the desktop affordance. Switching versions is a reading action; Duplicate and Download are management actions. Happy to change that if you'd rather ⋯ owned it outright.

## Verification note

I could not produce a live screenshot of the result. Published chrome (`mode: 'published'`) only comes from the worker — `server/server.js:323` hardcodes `mode: 'local'`, `/fork` is a worker-only route, and `preview.yml` needs Cloudflare secrets that a fork PR cannot see. The change is confirmed at the CSS level by diffing the overlay served by a patched local server against a stock one.

### DIFF
--- a/server/overlay.js
+++ b/server/overlay.js
@@ -471,8 +471,12 @@
   .tdoc-secondary-menu .tdoc-sec-sep { border-top: 1px solid #eee; margin: 4px 6px; }
 
   .tdoc-menu-wrap { position: relative; display: inline-block; }
-  /* Overflow ⋯ button shows on narrow viewports. */
-  .tdoc-bar .tdoc-secondary-toggle { display: none; padding: 6px 10px; }
+  /* Overflow ⋯ button is the single home for the secondary doc actions at
+     every width. Duplicate / Download / Save-as exist in the bar markup for
+     the export and fork chrome checks, but never show inline — the ⋯ menu
+     already carries the same actions. */
+  .tdoc-bar .tdoc-secondary-toggle { display: inline-flex; padding: 6px 10px; }
+  .tdoc-bar #tdoc-duplicate-btn, .tdoc-bar #tdoc-download-wrap, .tdoc-bar #tdoc-saveas-btn { display: none; }
   /* Identity chip — avatar + name (name hides on narrow). */
   .tdoc-chip { display: inline-flex; align-items: center; gap: 8px; padding: 3px 12px 3px 3px; background: #f0f1f4; border-radius: 999px; cursor: pointer; color: #1a1a1a; font: inherit; border: none; position: relative; }
   .tdoc-chip:hover { background: #e5e6ea; }
@@ -773,6 +777,8 @@
   /* Bar collapse breakpoints — tied to viewport width, not layout class.
      The bar progressively hides elements as the viewport tightens, so it
      stays elegant at every size.
+     ⋯ is present at every width and holds Duplicate / Download, so the bar
+     only ever sheds identity and breadcrumb detail as it tightens.
        ≥1100px: logo · slug · v · title ……………… identity · share · ⋯
        <1100px: logo ·      · v · title ……………… identity · share · ⋯  (slug hides)
        < 900px: logo ·      · v · title ……………… avatar   · share · ⋯  (name hides)
@@ -783,8 +789,6 @@
   @media (max-width: 900px) {
     .tdoc-chip .name { display: none; }
     .tdoc-chip { padding: 3px; }
-    .tdoc-bar #tdoc-duplicate-btn, .tdoc-bar #tdoc-download-wrap, .tdoc-bar #tdoc-saveas-btn { display: none; }
-    .tdoc-bar .tdoc-secondary-toggle { display: inline-flex; }
   }
   @media (max-width: 700px) {
     .tdoc-bar { padding: 0 8px; gap: 4px; }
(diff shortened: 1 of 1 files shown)


yayashuxue:
Thank you @unknowncici for this — genuinely 🙌. You spotted the right problem: the top bar was carrying more chrome than it needed, and Duplicate/Download shouldn't be fighting for space at every width. Your pure-CSS fix (collapsing them into the ⋯ overflow at all widths) was clean and exactly the right instinct.

We ended up folding your idea into a slightly larger pass — #240 — that also moves **Copy** into the ⋯ menu (as a single "Copy as Markdown" action) and unifies the whole thing to `Share + ⋯` at every width. Since #240 subsumes what this PR does (Duplicate/Download collapse), I'm closing this one to avoid the two touching the same CSS.

Really appreciate the contribution — hope to see more. 💚

PR #234, opened 2026-08-23
fix: let an agent comment status converge back to applied

### LINKED ISSUE #229: Agent comment status cannot converge back to applied within one version (question/partial → applied stays 🟡)
Once an agent marks a comment `partial` or `question`, a later `applied` reply at the same `at_version` never takes effect. The comment keeps the stale emoji forever, so the ✅/🟡/❓ verdict — the thing `/tdoc edit` relies on to tell a reviewer what happened — can lie permanently.

`question → applied` is the *normal* progression: the agent asks for clarification, the human answers, the agent applies it. That path is exactly the one that cannot converge.

## Cause

Two interacting pieces.

`eventEid` (worker.js:2119) gives status events an eid that **includes the kind**:

```js
case 'marked_applied':
case 'marked_open':
case 'deleted':
  return `${e.kind}:${e.at_version}`;
```

so `marked_applied:1` and `marked_open:1` are two independent slots.

`dedupEvents` refreshes each slot's **value** with the last event, but emits it at the **position of the first occurrence**:

```js
const lastByEid = new Map();
for (const e of events) { if (e && e.eid) lastByEid.set(e.eid, e); }   // last value
...
  if (emitted.has(id)) continue;
  emitted.add(id);
  out.push(lastByEid.get(id));                                          // first position
```

So `applied(t1) → open(t2) → applied(t3)` folds to `[applied(t3), open(t2)]`. The newest verdict is carried but re-seated ahead of the older one, and `open` ends up last, so it wins.

This is precisely the `add → remove → add` data-loss shape the comment above `eventEid` says was solved for reactions by **dropping the kind from the eid**. Status events never got the same treatment.

## Reproduction

Against the real functions in `worker/worker.js`, no network:

```js
appendEvent(c, {kind:'marked_applied', at_version:1, at:'04:10', agent_status:'applied'});
appendEvent(c, {kind:'marked_open',    at_version:1, at:'04:20', agent_status:'partial'});
appendEvent(c, {kind:'marked_applied', at_version:1, at:'04:30', agent_status:'applied'});
snapshotAt(c, 1)
```

```
event order after dedup:
    created         04:00
    marked_applied  04:30      <- newest value, oldest position
    marked_open     04:20      <- therefore last, therefore wins

expected : applied / ✅
actual   : open / 🟡
```

| scenario | result |
|---|---|
| current code, same version `[1,1,1]` | `open / 🟡` — bug |
| current code, bumped `at_version` `[1,1,2]` | `applied / ✅` — accidental workaround |
| eid patched to drop the kind | `applied / ✅` — converges |

## Seen in the wild

`tdoc.dev/d/tdoc-onboarding-redesign`, comment `c_1787445932434_a382237d`: three agent replies (applied → partial → applied). The newest reply carries `agent_status: "applied"`, and the server still reports the parent as `status: open`, `reactions: {🟡}`.

## Suggested fix

Give status events a kind-free eid, the same way reactions already converge:

```js
case 'marked_applied':
case 'marked_open':
  return `status:${e.at_version}`;
case 'deleted':
  return `${e.kind}:${e.at_version}`;
```

`deleted` should keep its own slot — it is terminal, not a toggle.

Worth adding a convergence case to `test/event-convergence.test.js` alongside the existing reaction one, since this is the same failure mode.

serenakeyitan:
## What & why

Fixes #229.

`marked_applied` and `marked_open` are a **toggle**, not two independent facts, but `eventEid` keyed them as `` `${e.kind}:${e.at_version}` `` — two separate slots. `dedupEvents` refreshes a slot's *value* with the last matching event but emits it at the *first* occurrence's position:

```js
for (const e of events) { if (e && e.eid) lastByEid.set(e.eid, e); }  // last value
...
  if (emitted.has(id)) continue;
  emitted.add(id);
  out.push(lastByEid.get(id));                                        // first position
```

So `applied(t1) → open(t2) → applied(t3)` folded to `[applied(t3), open(t2)]`: the newest verdict was carried, but re-seated *ahead* of the older one, so `open` ended up last and won.

The verdict could never converge back to applied within one version. `question → applied` is the normal progression — the agent asks, the human answers, the agent applies — so ✅/🟡/❓ could lie permanently, and that emoji is exactly what `/tdoc edit` relies on to tell a reviewer what happened.

The fix drops the kind from the eid, the same way reaction toggles already converge (the rationale comment above `eventEid` had described that fix for reactions but status never got it). `deleted` keeps its own slot — it is terminal, not a toggle.

### Migration is already handled

`marked_applied` / `marked_open` are in `DETERMINISTIC_EID_KINDS`, and `backfillEids` **recomputes** — not merely backfills — deterministic eids whose format changed. Existing stuck comments therefore self-heal on the next fold. There is a test for exactly this, driving events that carry the old `marked_applied:1` / `marked_open:1` eids.

Nothing anywhere parses or prefix-matches an eid — the only consumer is `const id = e.eid` as a Map key — so the format change is contained.

### Tests

Five cases added to `test/event-convergence.test.js`:

| case | fails before fix |
|---|---|
| status kinds share one eid slot; `deleted` does not; version still separates | yes |
| `applied → partial → applied` ends applied | yes |
| migration: legacy events under the old eid converge | yes |
| `question → applied` flips the verdict | no — guard |
| a later `marked_open` still reopens an applied comment | no — guard |

The last two pass either way and are direction guards, not regression proofs: the bug needs the *same* kind to repeat around the other (`[A, B, A]`). Verified by reverting only the eid change — `8 passed, 3 failed`.

## Checklist

- [x] `npm test` passes locally — `PASS — all 43 suite(s) green`.
- [x] Worker touched. `npm run test:all` was run; the Playwright suites (`ui`, `responsive`, `csp-xss`) skip loudly because Playwright is not installed here, and `publish.test.js` errors on `require('playwright')` at module top — **both reproduce identically on clean `main`**, so neither is from this change. The change is a pure string-returning switch in the event-fold layer with no DOM, CSP, or publish-path surface.
- [x] `SKILL.md` not edited.
- [x] No version change.
- [x] `worker/_worker.bundled.js` is not committed; CI regenerates it via `node bin/tdoc-bundle` on deploy, so no bundle refresh is needed here.

## Security note

Touches `worker/worker.js`, but only `eventEid`, which turns an event into a dedup key. It reads no request input, performs no auth decision, and emits no markup — `at_version` is already coerced by `coerceBodyVersion` upstream. No change to what a caller can write: the `/api/agent/reply` handler still gates on the upload token and on `parent_not_found` exactly as before. The effect is confined to which stored events collapse together during the fold.

### DIFF
--- a/test/event-convergence.test.js
+++ b/test/event-convergence.test.js
@@ -106,5 +106,75 @@ t('ORDERING: fold result is identical regardless of physical event order', () =>
   assert(inOrder.text === 'edited' && inOrder.status === 'applied');
 });
 
+
+// --- agent status is a toggle, and must converge back to applied (#229) ---
+// question/partial -> applied is the NORMAL progression: the agent asks, the
+// human answers, the agent applies. Under the old `<kind>:<at_version>` eid
+// that path could never converge, because marked_applied and marked_open sat
+// in two slots and dedupEvents re-seats each slot at its FIRST occurrence.
+
+t('marked_applied and marked_open share one eid slot (status is a toggle)', () => {
+  const applied = eventEid({ kind: 'marked_applied', at_version: 1 });
+  const open = eventEid({ kind: 'marked_open', at_version: 1 });
+  assert(applied === open,
+    `status kinds must share an eid, got ${applied} vs ${open}`);
+  // ...but not across versions, so snapshots stay immutable.
+  assert(eventEid({ kind: 'marked_open', at_version: 2 }) !== open,
+    'status eid must still include at_version');
+  // deleted is terminal, not a toggle — it keeps its own slot.
+  assert(eventEid({ kind: 'deleted', at_version: 1 }) !== open,
+    'deleted must not share the status slot');
+});
+
+t('CONVERGENCE: applied -> partial -> applied ends applied, not stuck open', () => {
+  const c = { id: 'c_1', created_in: 1, author: { login: 'u' }, text: 'q',
+    events: [{ kind: 'created', at_version: 1, at: 't0', eid: 'created:1' }] };
+  appendEvent(c, { kind: 'marked_applied', at_version: 1, at: 't1', applied_in: 1, by: 'claude', agent_status: 'applied' });
+  appendEvent(c, { kind: 'marked_open', at_version: 1, at: 't2', by: 'claude', agent_status: 'partial' });
+  appendEvent(c, { kind: 'marked_applied', at_version: 1, at: 't3', applied_in: 1, by: 'claude', agent_status: 'applied' });
+  const snap = snapshotAt(c, 1);
+  assert(snap.status === 'applied', `status stuck at "${snap.status}" after re-applying`);
+  assert(Object.keys(snap.reactions || {}).join('') === '✅',
+    `emoji is ${JSON.stringify(Object.keys(snap.reactions || {}))}, expected ✅`);
+});
+
+t('CONVERGENCE: question -> applied flips the verdict within one version', () => {
+  const c = { id: 'c_2', created_in: 2, author: { login: 'u' }, text: 'q',
+    events: [{ kind: 'created', at_version: 2, at: 't0', eid: 'created:2' }] };
+  appendEvent(c, { kind: 'marked_open', at_version: 2, at: 't1', by: 'claude', agent_status: 'question' });
+  appendEvent(c, { kind: 'marked_applied', at_version: 2, at: 't2', applied_in: 2, by: 'claude', agent_status: 'applied' });
+  const snap = snapshotAt(c, 2);
+  assert(snap.status === 'applied', `question->applied left status "${snap.status}"`);
+});
+
+t('a later marked_open still reopens an applied comment', () => {
+  // The toggle must work in both directions — converging must not mean
+  // "applied always wins".
+  const c = { id: 'c_3', created_in: 1, author: { login: 'u' }, text: 'q',
+    events: [{ kind: 'created', at_version: 1, at: 't0', eid: 'created:1' }] };
+  appendEvent(c, { kind: 'marked_applied', at_version: 1, at: 't1', applied_in: 1, by: 'claude', agent_status: 'applied' });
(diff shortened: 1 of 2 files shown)


serenakeyitan:
Manual end-to-end verification, driving the real `/api/agent/reply` endpoint through `worker.fetch` with fake KV/R2/DO bindings (same harness style as `hosted-oob-behavior.test.js`) — not the extracted pure functions.

### A/B through the real endpoint

Sequence: publish a doc → human comment → agent replies `applied`, `partial`, `applied`.

**Current `main` (5a720e5):**
```
reply 1 applied (HTTP 200) : {"status":"applied","emoji":"✅"}
reply 2 partial (HTTP 200) : {"status":"open","emoji":"🟡"}
reply 3 applied (HTTP 200) : {"status":"open","emoji":"🟡"}
>>> STUCK at open / 🟡
```

**This PR:**
```
reply 1 applied (HTTP 200) : {"status":"applied","emoji":"✅"}
reply 2 partial (HTTP 200) : {"status":"open","emoji":"🟡"}
reply 3 applied (HTTP 200) : {"status":"applied","emoji":"✅"}
>>> converged to applied / ✅
```

Note reply 3 returns **HTTP 200** on `main` — the write is accepted and then silently discarded by the fold. That is why this was invisible.

### Self-heal, on real stored data

Built the stuck state with the old worker, then handed the **same** storage to the fixed worker and only read:

```
stored status events BEFORE (written by old worker):
    marked_applied -> eid "marked_applied:1"
    marked_open    -> eid "marked_open:1"
    marked_applied -> eid "marked_applied:1"

read back by OLD worker : {"status":"open","emoji":"🟡"}
read back by NEW worker : {"status":"applied","emoji":"✅"}   <- no new writes
```

### One clarification to the PR body

The body says backfillEids "recomputes stored events under the new format." To be precise: the recompute happens on the **in-memory copy at fold time**, and a read deliberately does not rewrite the record — the stored eids are still `marked_applied:1` after the fixed worker read them (verified above). That is intentional and matches the note at `worker.js:742` about folding a copy so a log-only check cannot persist an incidental backfill.

The practical effect is the same or better: every read heals, and the on-disk form migrates whenever the record is next written for some other reason. No backfill migration job is needed.

Issue #236, opened 2026-08-23
Onboarding marches hosted users through four Cloudflare steps they do not need


serenakeyitan:
A user who only wants to publish to `tdoc.dev` — the CLI's own default target — is walked through installing wrangler, logging into Cloudflare, claiming a workers.dev subdomain, and enabling R2. None of those are needed to publish to hosted tdoc.

This is slice **S1 + S2** of the approved onboarding redesign: https://tdoc.dev/d/tdoc-onboarding-redesign/v/2

## Two halves of one defect

**The machine-readable gate.** `bin/tdoc-doctor:147` computes readiness with a single conjunction, and five of its seven terms are Cloudflare:

```bash
if $node_ok && $wrangler_ok && $jq_ok && $cf_logged_in \
   && $cf_publish_token_ok && $cf_subdomain_ok && $cf_r2_enabled; then
  ready_to_publish=true
```

The doctor has no notion of a target platform, so it answers the Cloudflare question for everyone. On a machine that has never installed a Cloudflare tool:

```
$ TDOC_MOCK_NO_WRANGLER=1 bin/tdoc-doctor | jq "{ready_to_publish, missing_steps}"
{
  "ready_to_publish": false,
  "missing_steps": [
    { "id": "wrangler", "label": "Install wrangler CLI", "kind": "install" },
    { "id": "cf_login",  "label": "Log into Cloudflare",  "kind": "login"   }
  ]
}
```

Claim-subdomain and enable-R2 get appended once login succeeds.

**The prose that drives off it.** `ONBOARDING.md` Step 4 tells the agent to walk `missing_steps` in order, and Step 5 (publish) is gated on `ready_to_publish`. So an agent following that file *cannot reach a hosted publish at all* without first satisfying a BYOK checklist.

Meanwhile `bin/tdoc-publish:684` has routed the unflagged case to `first_time_setup_hosted` for a while. Whether a given user is dragged through Cloudflare depends on whether their agent trusts `FIRST-DOC.md` (which says hosted is the default) or `tdoc-doctor` (which is machine-readable and says they are not ready). A default that survives only when the agent guesses right is not a default.

Same root cause as #226: surfaces that predate the hosted default and never learned about it.

## What hosted actually needs

| dependency | hosted publish needs | `ready_to_publish` demands |
|---|---|---|
| Node 18+ | yes | yes |
| curl | yes | not checked |
| jq | yes (today) | yes |
| wrangler CLI | **no** | yes |
[...]

Issue #193, opened 2026-08-21
Hosted onboarding should use browser-to-CLI pairing instead of a second GitHub device login


yayashuxue:
## Problem

Hosted onboarding currently has two separate GitHub auth moments:

1. First local publish uses GitHub Device Flow in `tdoc-publish` so the CLI can call `/api/hosted/token` and save an account-scoped `upload_token` in `~/.tdoc/published.json`.
2. The website uses browser OAuth for `tdoc.dev/me`, commenting, and owner management, creating a separate `tdoc_sid` browser session.

This is technically expected, but from a user perspective it looks like they have to log in twice during setup.

## Preferred direction

Do not expose the real hosted `upload_token` in browser UI or ask users to paste it. That token is a bearer credential and would likely end up in clipboards, terminal history, chats, or agent transcripts.

Instead, add a short-lived browser-to-CLI pairing flow:

- User signs in once on `tdoc.dev`.
- Website offers “Connect CLI” and shows a short-lived one-time code or copyable command, e.g. `tdoc login <code>`.
- CLI exchanges that one-time code with the hosted worker for the account-scoped upload token.
- CLI writes `~/.tdoc/published.json` exactly as today.

## Notes

Current implementation references:

- `bin/tdoc-publish`: `hosted_github_signin()` uses Device Flow and `first_time_setup_hosted()` calls `/api/hosted/token`.
- `worker/worker.js`: `/api/hosted/token` requires a signed-in session and mints the hosted account token.
- Browser OAuth now handles website sessions independently via `tdoc_sid`.

Acceptance shape:

- First hosted publish can be authorized from an existing browser session without approving GitHub Device Flow separately.
- The real upload token is never displayed in browser UI.
- Pairing codes are one-time, short-lived, scoped to the signed-in GitHub account, and invalidated after exchange.
- Existing Device Flow remains as fallback for headless/remote CLI environments.

serenakeyitan:
Evidence from manual testing of the onboarding work (#236), which kept walking into this.

**The cheap workaround does not exist.** GitHub returns `verification_uri_complete: null` on every `/api/auth/device/start` — verified against the live host:

```
$ curl -s -X POST https://tdoc.dev/api/auth/device/start -d "{}" | jq
{ "verification_uri": "https://github.com/login/device",
  "verification_uri_complete": null }
```

[...]

PR #239, opened 2026-08-23
/tdoc new publishes to tdoc.dev and never hands back localhost

serenakeyitan:
## What & why

Refs #236. Slice **S4 + S5 + S7** of the approved redesign: https://tdoc.dev/d/tdoc-onboarding-redesign/v/3

> **Stacked on #237.** Base is `feat/hosted-aware-doctor`, so the diff here is only this slice. Retarget to `main` once #237 lands. The tip of this branch is the full flow, which is what to check out for end-to-end testing.

A first doc ended at `http://localhost:7878` — a URL that cannot be shared, dies with the server, and means nothing to someone who has never seen tdoc. Publishing was a second thing the user had to know to ask for.

## Where a doc goes now

| The user said | Destination | Handed back |
|---|---|---|
| nothing about hosting | **hosted tdoc.dev** | `https://tdoc.dev/d/<slug>/v/1` |
| "self-host it" / "my own Cloudflare / Vercel" | their own worker | `<worker>.workers.dev` — still public, **not localhost** |
| "keep it local" / "don't upload it anywhere" | local only | `http://localhost:7878/…` |

**The localhost rule** is now stated as an invariant: never hand over a localhost URL unless it was asked for — not as a fallback, not when a sign-in did not finish. A publish that cannot complete says so and leaves the doc in `$TDOC_DIR`; it does not substitute a local URL for the promised link. Asking to self-host is **not** asking for localhost.

## Sign-in moves ahead of generation

`tdoc-publish --signin-only` mints the credential without needing a doc, so the skill starts it in the background and writes the HTML while the human approves — instead of interrupting them at the moment the link was about to appear. It is an instant no-op when already signed in, which is every run after the first.

The device flow now **opens the browser itself**, on GitHub's `verification_uri_complete`, so there is nothing to copy or type.

One thing worth reviewing closely: the auto-open is deliberately **not** gated on `[ -t 2 ]`. I wrote it that way first and it was wrong — an agent runs this with stderr piped, so a tty check would mean the browser never opens in the exact situation the feature exists for. It is gated on "is there a human desktop here" instead: off under `CI`, off over SSH, off without a display on Linux, off with `TDOC_NO_BROWSER=1`. The printed URL and code remain the fallback, and only a `https://github.com/*` URL is ever handed to the opener.

## Not touched, on purpose

`bin/tdoc-new` still defaults to `PUBLISH=0`. `/document-release`, `/retro`, `/investigate`, `/cso` and `/qa-only` all call it — flipping the default there would auto-upload every security audit and retro to tdoc.dev. There is now a test pinning it.

## Tests

Six cases added to `cli.test.js`; three fail on the pre-change tree (`34 passed, 3 failed`), three are guards on behaviour that must not regress.

Also ran a full user-journey simulation against a stub hosted service — **17/17**:

```
[Flow B1] agent-first, brand new machine   sign-in exits 0 · browser opened on the pre-filled URL ·
                                           token stored once · publish needs no second sign-in ·
                                           link is not localhost · no second browser open
[Flow B2] second doc, same machine         sign-in is an instant no-op · no browser · publish works
[Flow A]  site-first                       identical once the line is pasted
[Flow C]  "keep it local"                  no sign-in triggered, no browser, no token minted
[Flow D]  "self-host it"                   defers to publish-time setup, no hosted token minted behind their back
[Flow E]  CI and SSH                       browser suppressed, sign-in still completes
```

The first run of that simulation failed 4/17 — my stub had no `/api/upload`. Harness gap, not a code bug; fixed and re-run.

## Checklist

- [x] `npm test` — `PASS — all 43 suite(s) green`.
- [x] `onboarding.test.js` — `15 passed, 0 failed`.
- [x] Nothing browser-facing or worker-side; Playwright suites do not apply.
- [x] `SKILL.md` copied to `skills/tdoc/SKILL.md`, with a test that pins them identical.
- [x] No version change.

## Security note

`bin/tdoc-publish` handles the upload token. `--signin-only` reuses `first_time_setup_hosted` unchanged — same device flow, same mint, same `chmod 600` on `~/.tdoc/published.json`; it only stops before uploading, and no-ops when a config already exists, so it cannot mint a second credential or overwrite one. The new browser-open takes the URL from the host's device-flow response and is matched against `https://github.com/*` before being passed to `open`/`xdg-open`, so a hostile or malformed `verification_uri` cannot become an arbitrary command or a non-GitHub destination; it is best-effort and failures are swallowed. No token is ever passed to the opener or printed.

## Not merging

Opened for manual testing, per request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,16 @@ file and `.claude-plugin/plugin.json`.
 
 ### Fixed
 
+- **The project mark keeps a white field in both themes, and its hover
+  highlight is centred.** The SVG that replaced the raster was fully
+  transparent, so the dinosaur read as a see-through outline. The mark is now
+  ink on an opaque white field, the same look as the landing hero. Because it
+  carries its own field it is restored like a photograph in dark mode instead
+  of inverting with the page — inverting turned the field black, which is the
+  bar's own colour, so the drawing still read as see-through. Separately, the
+  bar mark's 24px logo sat flush left in its 32px button, so the hover
+  highlight landed 8px off to the right of the drawing; the button now centres
+  its content. The worker's inlined copy of the asset was re-synced. `#161`.
 - **`tdoc-agent-reply` fails loudly.** A rejected reply used to exit 0, so a
   comment that was never answered looked answered. `curl -sS` exits 0 on HTTP
   4xx/5xx, and the server also reports rejections as a 200 body with an
--- a/FIRST-DOC.md
+++ b/FIRST-DOC.md
@@ -10,17 +10,27 @@ Follow [ONBOARDING.md](ONBOARDING.md) end-to-end first. Do not skip it and do
 not ask the human which runtime they use; that file tells you how to install
 yourself and how to authenticate.
 
-Publishing is hosted by default. If the CLI asks the human to sign in with
+Publishing is hosted by default and needs no Cloudflare account, no card, and
+nothing to click in a dashboard. If the CLI asks the human to sign in with
 GitHub, print the code and the URL and wait for them. Do not ask them for a
 token; there is nothing for them to paste.
 
-**If the human said to publish to their own Cloudflare** — the line they pasted
-will say so — do that instead, and do not sign them into the hosted service.
-ONBOARDING.md step 4 covers it: install wrangler, `wrangler login`, then two
-things only they can click in a browser (claim a workers.dev subdomain, enable
-R2). Print each URL, say what to click, and wait for them to confirm before
-re-probing with `bin/tdoc-doctor`. It is about five minutes the first time and
-free. Their account, their bill, no card.
+Do not install wrangler, and do not send them to the Cloudflare dashboard. If
+`bin/tdoc-doctor` is asking for either, read its `.target` field first — on the
+hosted default it never should.
+
+<details>
+<summary><b>Only if the human said to publish to their own Cloudflare</b> — the
+line they pasted will say so</summary>
+
+Do that instead, and do not sign them into the hosted service. The self-hosting
+appendix at the end of ONBOARDING.md covers it: install wrangler,
+`wrangler login`, then two things only they can click in a browser (claim a
+workers.dev subdomain, enable R2). Print each URL, say what to click, and wait
+for them to confirm before re-probing with `bin/tdoc-doctor --platform
+cloudflare`. It is about five minutes the first time and free. Their account,
+their bill, no card.
+</details>
 
 ## Step 2 — make the doc
 
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -7,10 +7,10 @@
 `tdoc` is a Claude Code skill that gives the user prompt-native HTML documents with text- and artifact-anchored comments. After install + onboarding, the user can:
 
 - `/tdoc new <prompt>` → generate a commentable HTML doc
-- `/tdoc publish <slug>` → publish (default target is hosted tdoc.dev; first publish signs in with GitHub)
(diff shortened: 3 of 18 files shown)


yayashuxue:
make sure we still have the localhost version for BYOK users.
For hosted users, yes we want minimal onboarding frictions.

serenakeyitan:
Good catch to raise it — and worth being precise about what actually changed, because less changed than the diff makes it look.

**Nothing about the local server was removed.** `server/` is a zero diff in this PR, and `/tdoc serve`, `/tdoc stop`, `TDOC_PORT` and the port-conflict handling are all untouched. `bin/tdoc-new` is also unchanged and still defaults to `PUBLISH=0`, so `/document-release`, `/retro`, `/investigate`, `/cso` and `/qa-only` keep behaving exactly as before (there is now a test pinning that, since it would have been a quiet way to start auto-uploading every security audit).

The one real change for a self-hosting user is narrow: `/tdoc new` now publishes to their own worker and reports the worker URL, instead of opening a localhost preview first. The deliverable is a public URL either way — asking to self-host is asking to own the infrastructure, not to keep the doc off the internet.

I have pushed 66650bd to say that plainly in the skill, in both places it could be misread:

> **Local preview stays available to self-hosting users** — `/tdoc serve` and `http://localhost:7878` are unchanged, and iterating locally before pushing to your own worker is a perfectly good loop. That is an *authoring* step the user can ask for at any time; it does not change what gets handed over at the end, which is still the worker URL.

and next to the rule itself:

> This rule is about **what you hand over**, not about the local server, which is untouched.

One data point on the "BYOK needs local iteration because deploying is expensive" intuition, since I checked it rather than assumed: self-host publish only redeploys the worker when the **runtime bundle hash changes**, and `TDOC_SKIP_WORKER_DEPLOY=1` skips even that. A plain doc revision is just an upload, same as hosted. So the cost argument is weaker than it looks — but the workflow-familiarity argument stands on its own, which is why local preview stays and is now documented rather than implied.

Happy to go further if you want the self-host path to actually *default* to a local preview and publish only on request. That is a real product call rather than a bug, so I would rather have it be a deliberate decision than something I quietly pick.

serenakeyitan:
@yayashuxue one thing in here needs your call, because it runs against a decision you already made deliberately.

This PR has the CLI **open the browser** when the device flow starts. #186 removed exactly that from the web dialog, and your reasoning was specific:

> The popup was the mistake: on desktop it lands, so the dialog throws you into a GitHub tab the instant it appears — before you have read the code or understood what is happening. It reads as a hijack, not a sign-in.

My read is that the reasoning is about a browser tab and does not transfer to a terminal: there is no tab to lose, the poll loop is a local process rather than a page that can be navigated away, Safari activation windows do not apply, and the person just typed a command asking for a doc — a browser opening is a response to that, not an ambush. `signin-github-tab.test.js` guards the dialog and is untouched here.

But it is your call, not mine, and I would rather ask than quietly do the opposite of #186. Three options: keep it, drop it and just print the URL prominently, or put it behind a flag that is off by default. Happy with any of them.

Two other things from this round you may want to know about:

**`?user_code=` does not work.** I added it, claimed in the PR body that it pre-fills the code, and was wrong — I inferred it from GitHub carrying the parameter through its first redirect (`return_to=...user_code%3D...`). Walking it by hand shows a signed-out visitor hits the account chooser first and the parameter is gone by the time the code page renders. Reverted in 5947dd4, which lands back on exactly what #186 concluded. The code still has to be typed; the message now says it is only shown in the terminal, which was the original complaint.

**#193 is the real answer.** Manual testing kept running into "why do I have to type a code at all", which is what your pairing issue already describes. I have added the evidence there rather than trying to solve it in this PR.

PR #279, opened 2026-08-24
The owner's CLI can read their own private doc

### LINKED ISSUE #278: The owner cannot pull comments from their own private doc — /api/comments ignores the CLI token
### What happened

Continuing the end-to-end run on a fresh no-jq machine. After #276's fix `tdoc-pull` runs, and immediately hits a second wall:

```
$ tdoc-pull what-ai-knows-serena
tdoc-pull: worker did not return a comment array (network error or bad slug?). Local comments.json left untouched.
{"error":"access_denied"}
```

The doc is owned by the account whose token is sitting in `~/.tdoc/published.json`. It is `visibility: private`, `history_visibility: owner` — which is exactly what `FIRST-DOC.md` Step 7 tells the agent to publish every new user's first doc as.

### The token is valid; the read gate just never looks at it

```
# anonymous, private doc
$ curl "https://tdoc.dev/api/comments?slug=<slug>&version=all"
{"error":"access_denied"}

# the OWNER's upload_token on the same request
$ curl -H "Authorization: Bearer $TOKEN" "https://tdoc.dev/api/comments?slug=<slug>&version=all"
{"error":"access_denied"}

# the same token on /api/upload — 400 for the empty body, not 401
$ curl -X POST -H "Authorization: Bearer $TOKEN" -d '{}' https://tdoc.dev/api/upload
HTTP 400

# control: a public doc, anonymous
$ curl "https://tdoc.dev/api/comments?slug=tdoc-onboarding-redesign&version=all"
[{"id":"c_...","author":{"login":"serenakeyitan",...
```

`GET /api/comments` gates on `enforceDocAccess`, which resolves identity only through `getSession(env, req)` — a browser cookie. A CLI carrying a hosted account token is indistinguishable from an anonymous visitor, so the owner is denied their own comments.

### Why it matters

`/tdoc pull` is the documented pre-step to `/tdoc edit`: pull the comments people left, regenerate from them. On a private doc that loop cannot start. Since Step 7 makes private the first thing a new user publishes, the first doc they make is the one they cannot iterate on.

`tdoc-unpublish` and `tdoc-agent-reply` are unaffected — both go through upload-token-authenticated routes. It is specifically the read path.

### Suggested fix

`requireUploadAuth` already resolves a token to an actor via `hostedTokenActor`, and `requireDocWriteAccess` already compares `actor.account_id` against `meta.hosted.account_id`. The read gate can reuse both: if the request carries a Bearer token that resolves to the doc's owning account, treat it as the owner for `canReadDoc` / `canSeeHistory`.

Worth checking whether the same blindness affects any other read route that a CLI calls.

Found during the end-to-end onboarding run; separate from #276, which was about the jq dependency layer.

serenakeyitan:
Fixes #278. Found in the end-to-end onboarding run, immediately after #277 made `tdoc-pull` runnable at all.

### The bug

```
$ tdoc-pull <slug>
{"error":"access_denied"}
```

on a doc owned by the account whose token is sitting in `~/.tdoc/published.json`.

tdoc has two ways to prove who you are: a browser cookie, and the account token the CLI keeps. `enforceDocAccess` only ever knew about the first — it resolved identity through `getSession(env, req)` and nothing else — so a terminal holding a perfectly valid token looked exactly like an anonymous visitor.

The token was not the problem. The same token on `/api/upload` answers `400` for an empty body, not `401`:

```
$ curl -H "Authorization: Bearer $TOKEN" ".../api/comments?slug=<slug>&version=all"
{"error":"access_denied"}
$ curl -X POST -H "Authorization: Bearer $TOKEN" -d '{}' .../api/upload
HTTP 400
```

It was recognised there and never consulted here.

### Why it closes the loop the product is built on

`/tdoc pull` is the documented pre-step to `/tdoc edit`: pull what people left, regenerate from it. `FIRST-DOC.md` Step 7 publishes every new user's first doc as `private`. So the first doc anyone made was the one they could not iterate on.

### Server

`enforceDocAccess` falls back to the Bearer token when the session cannot read. The token is resolved exactly the way the upload path does it (`hostedTokenActor`, or the self-host admin token), and accepted only when `actor.account_id === meta.hosted.account_id` — the same equality `requireDocWriteAccess` already enforces for writes. **Reading over a token therefore grants precisely the set of docs that token can already overwrite, and nothing else.**

Two deliberate restraints:

- **It does not synthesise a session.** The caller keeps the real (possibly null) one, so an identity no browser ever authenticated cannot be rendered into a page or attributed to a comment. A test pins this, and a test pins that the token still cannot POST a comment.
- **The version list follows the same owner** (`gate.ownerToken || canSeeHistory(...)`), or the CLI would open a doc that claims to have one version.

### Client

`tdoc-pull` sends the token. It was the only CLI request that authenticated with nothing at all, with the token sitting in the same file it had just read `.base` out of. A config with no token still sends no header rather than an empty `Bearer` — a test pins that too, for self-host configs written before tokens existed.

A denial also gets its own message. `"network error or bad slug?"` sent people to look at their wifi; a denial has one likely cause and one fix, so it now says both.

### Testing

Server, against the real `worker.fetch` with the fake KV/R2/DO harness:

- the owner reads comments on their own private doc, and gets the comment an allowlisted reader left
- **a second account's token is still refused** — and still cannot write it either
- unissued and malformed bearers (`""`, `Bearer`, `../../etc/passwd`) are refused without a 500
- the owner sees all three versions over a token on a `history_visibility: owner` doc
- reading over a token renders `identity: null` and cannot post a comment
- an unlisted doc still needs no token — anonymous reads are not made harder anywhere

Client: the header is sent and the comments arrive; a token-less config does not become an empty `Bearer`; a denial is explained and leaves `comments.json` untouched.

Every one was checked against a deliberate reintroduction of the bug it claims to catch — eight mutations, eight failures, including the security-critical one (dropping the `account_id` comparison, caught by the stranger test).

`npm test` green (43 suites).

### Note

The worker half only takes effect on tdoc.dev once the Worker redeploys.

### DIFF
--- a/bin/tdoc-pull
+++ b/bin/tdoc-pull
@@ -69,9 +69,16 @@ fi
 # worker error must never clobber comments.json.
 if ! printf '%s' "$REMOTE" | json_is_array; then
   if printf '%s' "$REMOTE" | grep -q '"access_denied"'; then
-    echo "tdoc-pull: the worker refused to show '$SLUG' to this machine." >&2
-    echo "  This doc is private and the account token in $CONFIG_FILE does not own it." >&2
-    echo "  If it is yours, the token is stale — re-run: /tdoc publish $SLUG" >&2
+    # State the possibilities; do not pick one. The first live run of this
+    # message picked "your token does not own it" and was wrong: the token did
+    # own the doc, the worker was simply older than the fix that taught it to
+    # read tokens. Self-hosted workers can lag by any amount, so that case is
+    # permanent, not a launch-week detail.
+    echo "tdoc-pull: '$SLUG' is not readable by this machine ($BASE said access_denied)." >&2
+    echo "  The doc is private, so one of these is true:" >&2
+    echo "    - the account token in $CONFIG_FILE belongs to a different account" >&2
+    echo "    - this worker predates token-authenticated reads; redeploy it" >&2
+    echo "  If the doc is yours and the host is tdoc.dev, re-run: /tdoc publish $SLUG" >&2
   else
     echo "tdoc-pull: worker did not return a comment array (network error or bad slug?). Local comments.json left untouched." >&2
   fi
--- a/bin/tdoc-pull
+++ b/bin/tdoc-pull
@@ -53,12 +53,28 @@ mkdir -p "$(dirname "$OUT")"
 # Fetch the FULL history (all versions). Without version=all the worker returns
 # only the latest snapshot, which silently drops comments created on earlier
 # versions (root cause of "comments disappear on pull").
-REMOTE="$(curl -sS --connect-timeout 10 --max-time 30 "$BASE/api/comments?slug=$SLUG&version=all")"
+# Send the account token. This was the only CLI request that authenticated
+# with nothing at all, so on a private doc the worker saw an anonymous visitor
+# and denied the owner their own comments — and private is what FIRST-DOC.md
+# publishes every new user's first doc as. See #278.
+TOKEN="$(json_file_get "$CONFIG_FILE" upload_token)"
+if [ -n "$TOKEN" ]; then
+  REMOTE="$(curl -sS --connect-timeout 10 --max-time 30 \
+    -H "Authorization: Bearer $TOKEN" "$BASE/api/comments?slug=$SLUG&version=all")"
+else
+  REMOTE="$(curl -sS --connect-timeout 10 --max-time 30 "$BASE/api/comments?slug=$SLUG&version=all")"
+fi
 
 # Validate the response is a JSON array before touching local state. A transient
 # worker error must never clobber comments.json.
 if ! printf '%s' "$REMOTE" | json_is_array; then
-  echo "tdoc-pull: worker did not return a comment array (network error or bad slug?). Local comments.json left untouched." >&2
+  if printf '%s' "$REMOTE" | grep -q '"access_denied"'; then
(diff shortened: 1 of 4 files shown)


serenakeyitan:
Pushed a follow-up after running the branch against real tdoc.dev.

The client half works — `tdoc-pull` now sends the token. But the run produced a denial message that was **false**:

```
This doc is private and the account token in ~/.tdoc/published.json does not own it.
If it is yours, the token is stale — re-run: /tdoc publish <slug>
```

The token *did* own the doc. tdoc.dev's Worker simply had not been redeployed with the server half of this PR yet, so it still ignored the header. The message confidently named the wrong cause and sent the owner off to re-publish for nothing.

That is not a launch-week artifact — self-hosted workers lag by whatever their owner chooses, so "client newer than its worker" is a permanent shape, not a transient one. The message now states both possibilities and asserts neither:

```
tdoc-pull: '<slug>' is not readable by this machine (https://tdoc.dev said access_denied).
  The doc is private, so one of these is true:
    - the account token in ~/.tdoc/published.json belongs to a different account
    - this worker predates token-authenticated reads; redeploy it
  If the doc is yours and the host is tdoc.dev, re-run: /tdoc publish <slug>
```

Tests pin that both causes appear and that the message does not assert one as fact — both checked by removing each cause and watching the test fail.

PR #308, opened 2026-08-28
The tornado is five strokes

yayashuxue:
The mark is redrawn: five round-capped lines instead of a traced outline, in `assets/tdoc_logo.svg` and byte-identical in `worker.js`.

It keeps the two properties the previous mark established — **no background field, ink in `currentColor`** — so the bar, the neutral pages and the dark-mode invert behave exactly as before. What changed is the technique, so the test that guarded paint order and `evenodd` (both artifacts of tracing a raster) now guards `fill="none"` with a `currentColor` stroke.

The viewBox crops to the drawing plus 6%. The source art floats inside a 436 square; carrying that margin would leave the mark about half the box at the 24px bar and the 16px tab slot — the lesson the previous mark already recorded.

- `assets/favicon.svg` follows the same drawing and keeps its explicit light/dark swap, since a tab strip has no page invert to ride on.
- `assets/tdoc_logo_square.svg` / `tdoc_logo_square_dark.svg` are new — the filled square lockup as delivered, for an app icon or a slide, surfaces that bring their own field. A filled square reads as a tile against the bar, which is why it is not the in-app mark.
- The three 2500px rasters are regenerated from the vector: the Open Graph card from the square, the two off-page rasters transparent.

Checked in the real bar at both themes (light: dark ink; dark: inverted to white; the mark button's hover tint still shows through). `npm run test:all`: 52/52 green.

https://claude.ai/code/session_01DTsAHQ7Xmdghu9jStyx3WY


yayashuxue:
Deferring the redraw: feedback is that the original mark reads more like a tornado, so it stays. The home-screen icon gap this PR uncovered is real and lands separately.

PR #200, opened 2026-08-21
PR preview serves the homepage and /start

serenakeyitan:
Related to #199

PR preview already minted a `pr-<N>` host (#148) but `/` was the empty branded fallback — nothing seeded `tornado-doc`. Reviewing the homepage meant waiting for merge, or uploading to a personal Worker.

Each preview job now uploads the same payload production uses:

- `node bin/tdoc-landing-release` → slug **tornado-doc** so `/` is `landingResponse` (not a side slug like `tdoc-home`)
- `tdoc-start` as `/start`
- existing `conway-life` fixture for overlay chrome

Sticky **Open this** is the preview host root. Tutorial and overlay demo are extra links. Not tdoc.dev, not a personal Worker.

Preview KV is shared across aliases: the most recent seed is what `/` shows. The **code** (overlay) is still unique to this PR.

Verified on this PR's own preview:
- https://pr-200-tdoc-preview.jyshi1107.workers.dev/ — `tornado-doc`, `isLanding`, published chrome
- https://pr-200-tdoc-preview.jyshi1107.workers.dev/start — tutorial
- https://pr-200-tdoc-preview.jyshi1107.workers.dev/d/conway-life/v/2 — overlay demo

Use the `pr-200-` host, not the un-aliased `tdoc-preview.jyshi1107.workers.dev`.

Not merging until you say so.

PR #312, opened 2026-08-29
A preview can show the homepage, so two of them can be compared

yayashuxue:
A PR preview seeded only the demo doc, so its `/` answered with the **neutral fallback page**. Anything that touches site chrome — a mark, a bar control, a theme — could not be reviewed where it actually ships, and two versions of the homepage could not be put side by side at all.

- `preview.yml` uploads the newest `landing/tornado-doc/vN` out of the checkout, so a preview's `/` is *this branch's code carrying this branch's landing*.
- The sticky comment leads with the homepage; the demo doc stays as the second link, because a document is where the comment layer can be exercised. Its test asserted the opposite — that the link must not be the host root — which was right only while the root was the neutral page.
- `test/visual/hosted-compare.mjs` seeds the same doc and grows landing scenes (signed out, signed in), so **any two builds can be compared on the homepage locally**, without a deploy:

```bash
git worktree add /tmp/tdoc-main origin/main
SKILL_DIR=/tmp/tdoc-main OUT_DIR=/tmp/wb-old node bin/tdoc-bundle
SKILL_DIR=$PWD           OUT_DIR=/tmp/wb-new node bin/tdoc-bundle
node test/visual/hosted-compare.mjs /tmp/cmp /tmp/wb-old/_worker.bundled.js /tmp/wb-new/_worker.bundled.js
# → /tmp/cmp/h11b-landing.png is the homepage, old | new
```

`npm test` green; verified the harness renders the real landing (v44) with full chrome.

https://claude.ai/code/session_01DTsAHQ7Xmdghu9jStyx3WY

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -119,18 +119,19 @@ jobs:
           fi
 
       - name: Create tdoc-preview Worker if missing
-        working-directory: worker
-        run: |
-          set -euo pipefail
-          if ! wrangler deployments list --name tdoc-preview >/dev/null 2>&1; then
-            wrangler deploy --name tdoc-preview -c wrangler.preview.toml
-          fi
-
-      - name: Put preview-only upload token (never production)
         env:
           TDOC_PREVIEW_UPLOAD_TOKEN: ${{ secrets.TDOC_PREVIEW_UPLOAD_TOKEN }}
         working-directory: worker
         run: |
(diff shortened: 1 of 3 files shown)


yayashuxue:
Opened against the wrong branch: `feat/preview-landing` is the stale branch from the closed #200, so this PR showed that history rather than the change. Reopened as a fresh branch.

PR #217, opened 2026-08-22
Cross-origin iframe shell for author documents

yayashuxue:
Cross-origin iframe **shell** for author documents — chrome (bar/footer/composer/pins/cards) in the outer document, author content isolated in a same-origin, sandboxed `/frame` iframe (opaque origin). Fixes the "歪"/layout-instability class of bugs at the root: author CSS can no longer touch tdoc chrome. Same look/behavior as the overlay, reusing its CSS + markup via shared modules.

## Done (all 50 test suites green — offline + gated browser)
- **Comment features 1:1:** select-to-comment, pins, floating cards, replies, reactions (+ emoji picker), delete, publish modal, **re-anchor** ("move anchor" → rebind selection), **cluster pins** (co-located → count badge + popover), **agent-logo avatars**, **resolved chips**.
- **`?comment=` deep-links** (wide) and author **`data-tdoc-default-theme`**.
- **Perf:** cheap per-frame scroll reposition; cached anchor resolution (no re-walk per post).
- **Single-path cut:** every doc renders the shell — `?shell=1` gate and `injectOverlay` are gone; `cspHeader` gains `frame-src 'self'`; sign-in + onboarding moved into the shell chrome.
- **Worker parity (render path):** `bin/tdoc-bundle` inlines chrome + shell (as code, self-registering on `globalThis`) + chrome/probe (as client strings), `bundle_sha` covers them (fail-closed); new `/frame` route (`Sec-Fetch-Dest`-gated, access-gated, `frameCspHeader`); `serveDocVersion` renders the shell for docs, keeps the overlay for the homepage (SEO). New `worker-shell.test.js` (source + bundled-builder functional checks).
- Shared builder `server/shell.js`; `server/chrome.js`; `server/frame-probe.js`. Gated browser suites (`ui`/`csp-xss`/`responsive`) migrated to the frame/shell; obsolete reader-CSS assertions retired (model B); real gaps skipped **loudly** with reasons.

## Remaining (tracked)
- **Published-mode bar client wiring** (Share / Duplicate / Download / sign-in identity / owner-manage) — markup exists (`buildBar` mode:`published`); handlers not yet ported. Only renders in the worker, so it needs a **preview worker** to build+test.
- **Element/region commenting** (drag a marquee over a canvas/svg/image/`<pre>`; aid-stamped) — shell is text-selection only today.
- **Mobile comment drawer** + narrow-bar More-menu collapse.
- **`data-tdoc-copy`** author buttons (probe→shell clipboard bridge).
[...]

PR #326, opened 2026-08-30
Every path that writes a version bakes the reading template

yayashuxue:
> **Written up as a tdoc, with the diagrams:** [文档的样式从哪来](https://tdoc.dev/d/tdoc-styling-map/v/4)
> — the full timeline from `overlay.js` to today, flow diagrams of the three
> creation paths and the two serving paths, where each class of failure happens,
> and the three-layer split (defaults / recommendation / contract) this PR
> implements. Comment on it directly if a diagram is easier to argue with than a
> diff.
>
> Two documents to look at side by side, same content, both passing the
> validator before this PR: [A — follows the contract](https://tdoc.dev/d/layout-ab-normal/v/1)
> · [B — four deliberate width violations](https://tdoc.dev/d/layout-ab-adversarial/v/1)

Documents have carried their own reading template since #217. Three things had to be true for that to hold — the template reaches the document, every path that writes a version stamps it, and the validator judges rendering rather than taste. None of the three were.

## The template did not reach the document

`/frame` injected it only when the HTML contained no `id="tdoc-reader"` **and** no `max-width` anywhere. The second half was a proxy for "this document styles itself", and it punished exactly the documents that followed the authoring contract:

```
contract says   be responsive          →  @media (max-width: 520px)
contract says   the template owns type →  no font-family
fallback reads  "contains max-width"   →  "styles itself" → no template
result                                    Times New Roman, full width, no margins
```

**32 of 45 local documents were in that state; 23 of them for this reason.** `sierra-team-map` contains the string `max-width` exactly once, inside a breakpoint.

The proxy was never needed: the template is `:where()` zero-specificity throughout, so a document that does style itself wins every property it declares and is unaffected by the injection.

## Not every path stamped it

Baking lived in `bin/tdoc-new`, which SKILL.md frames as the entry point for *other* skills. Chat-driven `/tdoc new` and `/tdoc edit` are prose procedures the agent performs by hand — and prose cannot enforce anything.

| path | before | after |
|---|---|---|
| `bin/tdoc-new` | bakes | bakes |
| chat `/tdoc new` | bakes only if the agent happens to call that script | bakes |
| `/tdoc edit` | **never bakes** | bakes |

So a document created before creation-time baking could be edited any number of times and still ship without a template — it had no path to recover on its own. `tdoc-next/v3`, published 08-30, is one.

`bin/tdoc-write` is that path: one gateway that validates, bakes, writes `v<n>/index.html`, updates `meta.json`, and leaves an existing comment thread alone. `--version next` takes the highest of what `meta.json` records and what is on disk, because a hand-written version directory never reached `meta.json`. Both prose procedures now call it; `/tdoc new` went from 7 hand-performed steps to 5.

`/export` had the opposite bug — it stamped unconditionally, a patch from when documents were not self-contained — so every download of a baked document carried two identical 8.3KB blocks and a duplicate id.

## The validator judged taste

It rejected a `.note` whose margin was 22px where the style entry says 20px. It passed a `width: 1100px` inner container, a `calc(50% - 50vw)` full-bleed block, a 980px table with no scroll wrapper, and a `viewBox` 1400 wide.

Now split:

- **ERROR** — inert-under-CSP markup, missing viewport, a content root the probe cannot measure, an author `<footer>`, no body background, a figure clipped by its own viewBox or overflowing the column
- **NOTE** — house-style deviations: background colour, component values, global selectors, accent, table cell reset. Printed, and the document is still created.
- `--strict` fails on notes, for CI.

Enforcing taste is the wrong shape for a tool whose defaults are deliberately zero-specificity so the author can win.

## Also in here

- **`bin/tdoc-bake`** — the single implementation, and a migration: `--scan` reports, `--scan --apply` stamps. This is the one-time migration PLAN.md promised and never landed.
- **`data-tdoc-template="<sha8>"`** on the baked block. Each document freezes the template it was created with, so the generation should be readable from the document rather than inferred by comparing h1 sizes across documents.
- **SKILL.md** — its own "What to write" scaffold now passes its own validator (it was missing `body { background }`). 16 references to the deleted `overlay.js` corrected, including `Light mode only; the overlay does not currently support dark mode`, which contradicted both `chrome.css` and `authoring/style/technical.md`; and the `max-width: 720px` container guidance the validator rejects, replaced with the canvas number (672px).
- **`authoring/README.md`** — the precedence chain (user's words > declared style > recommendation > defaults), a three-line design plan before writing CSS, and a named list of looks to avoid. Borrowed from Claude's `artifact-design`. Also corrected "style/ and structure/ are empty on purpose"; both have shipped entries for a while.
- **`test/dark-invert-parity.test.js`** — the dark invert exists twice by necessity (shell document vs sandboxed frame) and had already drifted: `chrome.css` restores `.tdoc-emoji` to true colour, the probe's copy does not. Merging them means touching the Vite build and two injection points, so this pins them instead.
- Removed `body.tdoc-catalog` rules from `chrome.css` — nothing sets that class, and `reader.css` no longer reaches the shell.

## Verification

- 48/48 offline suites, including 7 new regression tests. One asserts the fallback can never key off `max-width` again.
- A document with a responsive breakpoint and no template gets one through `/frame`: **0 → 1**.
- `tdoc-write --version next` on a real document: baked, `meta.json` appended, earlier versions and `comments.json` untouched.
- 86 local versions rebaked; naked-document count across 49 documents **32 → 0**.

## Deliberately not in here

`style/*.md` still sets typography that duplicates the baked template (`default.md` sets `.wrap` font-size and line-height; four entries set an h1 size). Fixing that, trimming `reader.css`'s dead rules (the `.tdoc-bar` exclusion list and the `@media print` block hide chrome that no longer shares the document), and any narrowing of the floor all change how new documents look — so they belong together in one deliberate generation change, not spread across three. There are already two generations in circulation.


serenakeyitan:
@serenakeyitan review this pr

serenakeyitan:
Read this against the branch and ran the offline suite — 48/48 green. The `max-width` proxy was clearly the right thing to delete, and the `injectReaderCss` guard against double-stamping is a good catch. Three things from the read.

### 1. `background: transparent` passes the new structural check

The new ERROR only inspects the property *name*, never the value:

```python
if any(prop in {"background", "background-color"} for prop, _ in declarations):
    body_background = True
```

So a document that explicitly renders transparent passes, while one that omits the property fails — with an error message describing the first case:

| `body { background: … }` | result |
|---|---|
| `#fff` | exit 0, clean |
| `transparent` | **exit 0** (2 notes) |
| `none` | **exit 0** (2 notes) |
| `rgba(0,0,0,0)` | **exit 0** (2 notes) |
| *property omitted* | exit 1 — "body sets no background; the page renders transparent over the reader's ground" |

The three that pass are exactly the failure the error exists to name. Notes don't block, and `--strict` is CI-only, so nothing stops them. Looks like it wants the value checked too, not just the property.

### 2. `:where()` isn't quite "throughout", and that claim is load-bearing

Dropping the guard rests on "the template is `:where()` zero-specificity throughout, so a document that styles itself wins every property it declares". `reader.css` has three exceptions:

- **`:root`** — the 20+ design tokens, including `--td-ground` and `--td-ink`. Specificity `(0,1,0)`, not zero.
- `.tdoc-table-scroll` / `> table` — targets tdoc's own wrapper, so harmless in practice.
- `@media print` — `body { padding-*: 0 !important }`.

And the injection point is `body.replace(/<\/head>/i, rtag + '</head>')`, i.e. immediately *before* `</head>` and therefore *after* any author `<style>` in head. At equal specificity the later rule wins, so a legacy self-styled document that sets its own `:root` tokens would now have them overridden by the injection — the population this PR newly exposes.

Blast radius today looks small: nothing in `authoring/` or SKILL.md tells authors to set `--td-*`, so few documents likely do. But the invariant as written is stronger than the code supports, and it's the whole justification for removing the guard. `:where(:root)` would make the sentence true — including the one at the top of `reader.css` itself.

### 3. Latent: `$` in the replacement string (`bin/tdoc-bake:54`)

```js
html.replace(/<\/head>/i, `${tag}</head>`)
```

`tag` carries the entire template, and in a replacement string `$&`, `` $` ``, `$'` and `$1` are substitution patterns:

```
css = 'body{content:"$&"}'  →  <style id="tdoc-reader">body{content:"</head>"}</style>
```

`reader.css` has no `$` today, so this is latent rather than live. But `[class$="…"]` is ordinary CSS, so it's a plausible future footgun for a file explicitly meant to grow. A function replacement (`() => tag + '</head>'`) makes it inert. Three of the four sites predate this PR — only `tdoc-bake` is new here, so this is a "while you're in there".

---

Nit: `versionFiles`' comment says "newest slug first" but the sort is plain alphabetical ascending.

None of this is blocking — 1 is the only one I'd want before merge, and it's a one-liner. The ERROR/NOTE split and the single `tdoc-write` gateway are both clearly right.

<sub>Reviewed with Claude Code.</sub>

yayashuxue:
All three landed in `aa9333a`, with a regression test each.

**1 — the value, not just the property.** You're right that the message described a failure the check couldn't see. It now reads the value, and covers the modern slash syntax as well as the legacy comma one, since `hsl(0 0% 100% / 0)` slipped through my first pass at it:

| `body { background: … }` | before | now |
|---|---|---|
| `#fff` | 0 | 0 |
| `transparent` · `none` · `rgba(0,0,0,0)` | **0** | 1 |
| `hsl(0 0% 100% / 0)` · `oklch(… / 0%)` | **0** | 1 |
| *omitted* | 1 | 1 |

**2 — `:where(:root)`, and the claim rewritten to match.** This was the important one: the invariant *is* the justification for deleting the guard, and I stated it more strongly than the file supported. The token block is wrapped now. The two remaining non-zero rules are named in the file header instead of glossed — `.tdoc-table-scroll` targets tdoc's own wrapper, `@media print` only applies to printing — so the next person reads what is actually true rather than the sentence I wanted to be true.

Worth noting for the record: this makes the injection *safer* for exactly the population the PR newly exposes. Before, a legacy document that set its own `--td-*` would have had them silently overridden by the fallback.

**3 — callback at all four sites,** not just the new one. `bin/tdoc-bundle` already had this fix with a comment explaining why, so the codebase knew and the knowledge hadn't reached the other four. Test bakes a template containing `[class$="x"]` and `$&` and asserts both survive.

Nit fixed.

**One thing I took back out:** the `body.tdoc-catalog` removal from `chrome.css`. It is dead — nothing sets that class — but `server/runtime/` is a committed Vite build, so touching the stylesheet requires rebuilding and committing that too, and a locally-produced bundle isn't guaranteed to match what CI builds. It was this branch's only red check. It joins the deferred cleanup (`reader.css`'s `.tdoc-bar` exclusion list, the `@media print` block, `style/*.md`'s duplicate typography) — one deliberate generation change rather than three scattered ones.

48/48 green.

yayashuxue:
Correction to what I said about the `body.tdoc-catalog` removal — I claimed a locally-produced bundle isn't guaranteed to match what CI builds, and that was wrong. The `shell runtime` check *is* `npm ci && npm run build:shell && git diff --exit-code server/runtime`, so reproducibility is the premise the repo already runs on, not something I needed to be cautious about. I checked instead of assuming this time: a clean-tree rebuild reproduces the committed bundle byte for byte.

So it's back in, properly, as `f1cca8a`. The delta is only those rules — stripping `body.tdoc-catalog{…}` from the previous CSS bundle yields the new one exactly, 307 bytes, and the JS is unchanged apart from its content hash.

Why they were dead: `body.tdoc-catalog` existed when the reader template and the chrome shared one document, to stop the template from restyling the /me list. After the shell migration `reader.css` never reaches the shell, and nothing in `server/`, `shell/` or `worker/` sets the class.

That leaves the deferred list as genuinely one thing rather than one thing plus an excuse: `reader.css`'s `.tdoc-bar` exclusion list and `@media print` block, and `style/*.md`'s typography that duplicates the baked template. Those change how new documents look, so they want to be one deliberate generation change.

48/48, and the `shell runtime` check reproduces clean locally.

serenakeyitan (approved):
<h3>Breeze Summary</h3>

Follow-up on `aa9333a`, which answers the three findings from my read of this branch. All three are fixed at the source rather than papered over, each with a regression test, and the `body.tdoc-catalog` revert is the right call — a committed Vite build is not something to rebuild inside a PR about delivery. Re-read the branch and re-ran everything.

- **Finding 1 fixed.** `is_transparent()` now reads the value. Confirmed by running the validator: `transparent` · `none` · `rgba(0,0,0,0)` · `hsl(0 0% 100% / 0)` · `TRANSPARENT` all exit 1 now; `#fff` exits 0. `compact_css_value` lowercases and strips whitespace first, so `hsl(0 0% 100%/0)` and the spaced form both land on the same branch.
- **Finding 2 fixed.** `grep ':root' server/reader.css` returns exactly one selector and it is `:where(:root)`. The only non-`:where` rules left in the file are `.tdoc-table-scroll` (+ `> table`) and the `@media print` block — precisely the two the new header names, so the header is now a true statement about the file rather than an aspiration.
- **Finding 3 fixed at the four head-injection sites.** `bin/tdoc-bake:58`, `server/server.js:739`, `worker/worker.js:1496`, `worker/worker.js:3494` all pass a callback.
- **48/48 offline suites green** locally, including the three new tests. All five CI check-runs on `aa9333a` are green — the `shell runtime` check that the catalog removal broke is back to success.
- **Verified the revert restored the rules,** not just deleted the deletion: the four `body.tdoc-catalog :where(…)` rules are back in `server/chrome.css` byte-for-byte.

Merge. Not a 5 only because the `$` fix stops one line short of the injection site with the largest replacement payload — issue 1 below, which is the same defect this commit just fixed, two words away.

| Filename | Overview |
|---|---|
| `bin/tdoc-validate-template` | New `is_transparent()`; the structural body-background check now reads the value, and the error text names both failure shapes. |
| `server/reader.css` | Token block wrapped in `:where(:root)`; header now names the two deliberate non-zero-specificity rules instead of claiming "throughout". |
| `bin/tdoc-bake` | Callback replacement; `versionFiles` comment corrected to the ordering the sort actually produces. |
| `server/server.js` · `worker/worker.js` | Callback at the three remaining reader-CSS injection sites. |
| `server/chrome.css` | `body.tdoc-catalog` removal reverted — the branch's only red check. |
| `test/bake-reader.test.js` · `test/cli.test.js` | One regression test per finding; the `:where(:root)` one greps the stylesheet, so it holds for future edits too. |

<h3>Issues</h3>

**1. `server/server.js:748`, `worker/worker.js:3498` — the probe injection is the same bug, unfixed** *(nit, latent)*

Both still do ``body.replace('</body>', `${tag}\n</body>`)``, and a string first argument does not exempt the replacement from `$` patterns. `tag` here wraps `PROBE_JS` — 32KB of JavaScript, versus the CSS that motivated the fix. The argument in the commit message applies with more force: `$'` is what you get from any regex-in-a-string that anchors at the end, e.g. `const re = '^\s*$';`. Demonstrated on the actual code shape:

```js
> body.replace('</body>', `${tag}\n</body>`)
'<html><body>hi<script>const re=\'^x</html>;</script>\n</body></html>'   // $' expanded
> body.replace('</body>', () => `${tag}\n</body>`)
'<html><body>hi<script>const re=\'^x$\';</script>\n</body></html>'       // literal
```

`grep` confirms no `$&`, `` $` ``, `$'`, `$$` or `$<digit>` in `server/frame-probe.js` today, so this is latent exactly like the CSS one was — but the failure mode is worse: the probe silently breaks and stray markup lands inside a `<script>`. Two `() =>`.

**2. `bin/tdoc-validate-template` — hex zero-alpha still passes** *(nit)*

`is_transparent()` covers the functional notations but not the hex ones. Confirmed against the branch:

| `body { background: … }` | exit |
|---|---|
| `#ffffff00` | **0** |
| `#fff0` | **0** |
| `rgba(0,0,0,0)` | 1 |

8- and 4-digit hex is ordinary modern CSS and composites over the reader's ground identically. A `re.fullmatch(r"#(?:[0-9a-f]{3}0|[0-9a-f]{6}00)", v)` branch closes it. (`inherit` and `revert` are also absent from the keyword set, but those are genuinely rare on `body`.)

**3. `bin/tdoc-bake` — the migration has no re-stamp path, so fix 2 does not reach what this PR already rebaked** *(nit, operational)*

`bakeFile` returns early on `hasBlock(html)`, and `--scan` counts an already-stamped file as `present`. That idempotence is correct for the missing-template migration, but `aa9333a` changed the template, so the 86 versions the PR reports rebaking carry the bare `:root` generation — and a legacy document that sets its own `--td-*` is the population fix 2 exists to protect. `data-tdoc-template="<sha8>"` was added precisely so the generation is readable from the document; nothing reads it yet. A `--restamp` that replaces a block whose sha differs would make the stamp earn its keep. Fine to defer with the rest of the generation change, but worth deciding before the rebake is treated as done.

No blocking issues.


---

yayashuxue:
CodeQL caught something worth reading twice — in the test **this PR added**, and it is the same failure mode the PR is about.

`test/dark-invert-parity.test.js` existed to stop the two copies of the dark invert from drifting. It contained:

```js
if (!probe.includes(transform.replace(/ /g, ' ')))
```

`.replace(/ /g, ' ')` replaces a space with itself. An unfinished attempt at normalising whitespace, and the whitespace was the whole problem: `chrome.css` writes `filter: invert(1) hue-rotate(180deg)` and the frame's copy writes `filter:invert(1) hue-rotate(180deg)`. The assertion passed only because the substring it happened to check starts after the colon.

So the guard against a claim that outruns the code was itself a claim that outran the code.

Fixed in the latest commit: both sources are compared with whitespace removed, and the assertion covers the whole opening declaration block rather than one fragment. Verified by mutation rather than by reading it —

```
hue-rotate(180deg) → 170deg on one side   old test: passes    new test: fails
reverted                                   old test: passes    new test: passes
```

The other six open CodeQL alerts on the repo are pre-existing (`worker.js` sanitisation, stack-trace exposure) and untouched by this branch.

serenakeyitan:
<h3>Breeze Summary</h3>

Re-read the follow-up (`6cf1f64` + `5f062cc`) against the branch. All three findings from the previous pass landed with a regression test each, and the `body.tdoc-catalog` removal is back in with the rebuild done properly. The correction about bundle reproducibility is right — I checked it rather than taking it on faith.

- **Background value, not property name.** `transparent`, `none`, `initial`, `unset`, `rgba(0,0,0,0)`, `hsl(0 0% 100% / 0)`, `oklch(… / 0%)`, `color(… / 0)` all exit 1 now; `#fff` still exits 0. Both the legacy comma alpha and the modern slash alpha are covered, as claimed.
- **`:where(:root)`.** `server/reader.css` has no bare `:root` left, and the test pins it (`/^\s*:root\s*[,{]/`). Swept the whole file: the only non-zero-specificity rules are `.tdoc-table-scroll{,> table}` and the `@media print` block — exactly the two the new header names, so the header is now true rather than aspirational.
- **`$` in the replacement string.** All four `replace(/<\/head>/i, …)` sites take a callback (`bin/tdoc-bake:58`, `server/server.js:739`, `worker/worker.js:1496`, `worker/worker.js:3494`). The test bakes `[class$="x"]` plus a literal `$&` and asserts both survive.
- **The bundle.** Stripping the four `body.tdoc-catalog …` rules from the previous CSS yields the new file byte for byte — 307 bytes, confirmed by diff, not by eyeball. The JS is *identical* (same sha1), only its content-hash name moved. `grep -r tdoc-catalog` over the tree is empty.
- **Reproducibility.** Ran the `shell runtime` check locally on a clean tree: `npm ci && npm run build:shell && git diff --exit-code server/runtime` exits 0. The correction in the second comment is correct; the original caution was unnecessary.
- 48/48 offline suites green locally, and all five checks green on `5f062cc`.

Merge. Not a 5 only because the widened background check still has one hole of its own shape — a zero-alpha *hex* — which is a one-line addition, not a rethink.

| Filename | Overview |
|---|---|
| `bin/tdoc-validate-template` | New `is_transparent()`; the body-background ERROR now reads the value and its message names both failure shapes. |
| `server/reader.css` | Token block wrapped in `:where(:root)`; header states the two deliberate exceptions instead of glossing them. |
| `bin/tdoc-bake` | Callback replacement, with the reason inline; `versionFiles` comment now matches the sort. |
| `server/server.js`, `worker/worker.js` | Same callback fix at the three pre-existing injection sites. |
| `server/runtime/shell.*.{css,js}` | Rebuilt shell bundle; CSS −307 bytes, JS bytes unchanged. |
| `test/cli.test.js`, `test/bake-reader.test.js` | One regression test per finding: transparent values, bare `:root`, literal `$`. |

<h3>Issues</h3>

1. **`bin/tdoc-validate-template` — zero-alpha hex still passes (nit).** `is_transparent` enumerates keywords and colour *functions*, but `#rgba`/`#rrggbbaa` are neither. Verified on the branch with the same fixture the new test uses:

   | `body { background: … }` | exit |
   |---|---|
   | `transparent` · `rgba(0,0,0,0)` · `hsl(0 0% 100% / 0)` | 1 |
   | `#fff0` | **0** (2 notes) |
   | `#ffffff00` | **0** (2 notes) |

   Same failure the error exists to name, same "notes don't block" consequence — just written in hex. A `re.fullmatch(r"#[0-9a-f]{3,4}|#[0-9a-f]{6,8}", v)` branch that reads the trailing alpha nibble/byte closes it.

2. **`bin/tdoc-validate-template` — `any()` over all body rules ignores cascade order (nit).** `body_background` is true if *any* body declaration anywhere is non-transparent, so `body{background:#fff} … body{background:transparent}` passes while rendering transparent. Contrived enough that I would not hold the PR for it, but it is the same "the check does not model what the browser does" shape as finding 1.

3. **`server/reader.css` — one line of the new header is slightly generous (nit).** "neither competes with an author for a property they would set" is true of `.tdoc-table-scroll`, but `@media print { body { padding-*: 0 !important } }` does beat an author's print padding. The header does say the block is print-only, so a reader gets there; the clause just claims a bit more than the rule below it.

Nothing here blocks. The three fixes are the right fixes, and finding 2 of the previous pass — the `:where(:root)` one — came back stronger than I asked for: naming the surviving exceptions in the file is more durable than the invariant sentence alone.

---

serenakeyitan:
<h3>Breeze Summary</h3>

Checked `7217cdf` on its own. The claim is right and I confirmed it the same way you did — by mutation, not by reading. The old assertion was inert; the new one is not. The CodeQL alert on the file is gone, and the six that remain really are pre-existing.

- **The old test was inert, as described.** Changing `hue-rotate(180deg)` → `170deg` in `server/chrome.css:514`: old test **passes**, new test **fails**. Changing `background: #fff` → `#000` in the same block: old **passes**, new **fails**. Same 180→170 mutation applied to the probe's copy: new test **fails**. So `.replace(/ /g, ' ')` was not the only reason it passed — the substring boundary was, and the fix addresses that rather than the lint.
- **`squash` is the right shape for the two sources.** `chrome.css:511-515` writes the block over five lines; `frame-probe.js:469` writes it as one minified string. Stripping all whitespace makes them the same text, and the assertion now spans `color-scheme:dark;background:#fff;filter:invert(1) hue-rotate(180deg);` rather than one fragment of it.
- **Dropping the probe's `color-scheme:light` rule fails the new test.** Test 3's `src.includes('color-scheme:light')` is equivalent to the old regex once both sides are squashed, so nothing was lost in that rewrite.
- **CodeQL.** No open alert on `test/dark-invert-parity.test.js`. The six open ones are all `ref=refs/heads/main`, created 2026-06-28 (×4), 08-13 and 08-28 — all older than this branch. One small correction to your comment: five are in `worker/worker.js`, the sixth is `js/stack-trace-exposure` in `test/visual/hosted-compare.mjs:75`.
- 48/48 offline suites green locally on `7217cdf`; all five checks green on the PR.

Merge. Not a 5 for the same reason as last pass — a short list of one-line nits, none of which change behaviour, plus one new one below of exactly the shape this commit is about.

| Filename | Overview |
|---|---|
| `test/dark-invert-parity.test.js` | `squash()` applied to both sources; page-invert assertion widened from a fragment to the whole opening declaration block; `color-scheme` regex replaced with an equivalent squashed substring. |

<h3>Issues</h3>

1. **`test/dark-invert-parity.test.js` — the *restore* rule's transform is still compared as a selector only (nit).** Test 1 now pins the page invert's value, but test 2 checks only that each source contains `img:not([data-tdoc-dark="invert"])` and friends — not what filter those selectors carry. Verified: changing the probe's restore rule from `filter:invert(1) hue-rotate(180deg)` to `170deg` leaves the suite **green**. That is the same "the guard checks a fragment" failure the commit exists to fix, one rule further down; the fix is to give test 2 the same treatment test 1 just got.

2. **`test/dark-invert-parity.test.js` — squashing the whole file can't tell a live rule from a commented-out one (nit).** `squash` runs over the entire stylesheet, so wrapping `chrome.css:511-515` in `/* … */` still leaves the substring present and the test **passes** with the invert disabled. Contrived enough that I would not hold anything for it, but worth a line in the header comment so the next person knows the guard is textual.

3. **Leftover nits from the previous pass, unchanged by this commit** — restating so they are not lost between comments: `is_transparent` still misses zero-alpha hex (`#fff0`, `#ffffff00` exit 0); `body_background`'s `any()` still ignores cascade order; and the `reader.css` header's "neither competes with an author for a property they would set" is still slightly generous about the `@media print` block. All one-liners, none blocking.

The mutation table in your comment matches what I got independently. Worth saying plainly: a test that was doing nothing, in a PR whose subject is claims that outrun the code, is the most useful kind of thing for CodeQL to have caught here.


---

PR #116, opened 2026-08-15
Nag BYOK installs when origin/main is ahead

yayashuxue:
## What & why

BYOK overlay updates were invisible until someone happened to publish. GitHub releases lag (`v0.9.0` while main kept shipping), and the skill only checked those once a day. CLI + skill runs now compare this checkout to `origin/main` and immediately point at `/tdoc update --yes`.

Does not auto-deploy other people’s Workers. tdoc.dev CD is a separate thread.

## Checklist

- [x] `npm test` passes locally (offline suite; `test/cli.test.js` covers the nag).
- [x] Browser-facing overlay/worker unchanged — n/a for `test:all`.
- [x] `SKILL.md` copied to `skills/tdoc/SKILL.md`.
- [x] Version unchanged.

## Test plan

- [ ] On a skill checkout behind `origin/main`, run `tdoc-publish` / `tdoc-pull` / `tdoc-new`: stderr nags and stdout has `TDOC_UPDATE_AVAILABLE`.
- [ ] Same checkout: `tdoc-doctor | jq .update` shows `ok: false` and a `tdoc-update --yes` cmd (stdout stays valid JSON).
- [ ] Skill preamble prints `TDOC_UPDATE_AVAILABLE`; agent offers `/tdoc update --yes` before the rest of the work.
- [ ] Diverged checkout prints `TDOC_UPDATE_DIVERGED` and does not recommend `--yes`.
- [ ] `TDOC_SKIP_UPDATE_CHECK=1` is silent (used by hermetic CLI tests).

## Security note

n/a for auth/token paths. The nag is read-only `git fetch` with a stall timeout and never fails the caller. `tdoc-doctor | jq` stays valid because doctor uses `--json` and discards nag stderr.

Made with [Cursor](https://cursor.com)

PR #341, opened 2026-09-01
Storing a document version becomes one function

yayashuxue:
> Follow-up to #326, which fixed the read side (the render fallback no longer guesses). This is the write side: the server stops depending on the client having done the right thing.
>
> Context and the full RCA, as a tdoc: [文档的样式从哪来](https://tdoc.dev/d/tdoc-styling-map/v/7) — §10 (the three axes) and §13–15 (the self-update chronicle, the performance accounting, the client-version posture) are the sections this PR implements.

## The gap

The worker had no function named "store a version". Three paths stored documents, each assembling the invariants itself:

| writer | aid stamp | bake | sha |
|---|---|---|---|
| `/api/upload` | ✓ | ✗ | ✗ |
| browser Save (DO, #329) | ✓ | ✗ | ✗ |
| duplicate | ✓ | ✗ | ✗ |

So a version saved in the browser landed in R2 without the reading template — self-containment depended on which client produced the bytes, and the client's self-update machinery has failed silently in seven distinct ways across five generations (#116 → #248 → #268 → #332 → #334). Correctness cannot ride on that.

## The fix

`prepareDocVersion()` — every document write goes through it:

1. **Bake** the reading template, generation-stamped (`data-tdoc-template="<sha8>"`). No-op when the block is already there, and on the unbundled worker. Storage becomes self-contained regardless of client version; the render fallback (#326) keeps covering pre-existing storage and new writes never need it.
2. **Stamp aids**, as before.
3. **Hash the exact stored bytes.** Version entries record `sha`; the upload response returns it. This is what the next PR's read-side gating compares against (one HEAD instead of re-downloading).

Widgets deliberately do **not** come through it — they are sandboxed islands, and the template does not belong in them. A test pins that.

Plus tier-1 client visibility: `tdoc-publish` sends `X-Tdoc-Client`, the version entry records it. First time the wild version distribution is observable — at the one place guaranteed to see every publish.

## Performance

Bake cost moves to write time (once per version). The read path gains nothing and loses nothing: its per-request work is unchanged (R2 GET + probe injection dominate; the fallback's presence check is a microsecond substring scan that now almost never fires positive).

## Verification

- New `write-invariants.test.js`: all three writers through the real fetch handler against fake bindings — baked + stamped + `sha` recorded; pre-baked doc not double-baked; widgets not baked; `X-Tdoc-Client` recorded.
- 51/51 offline suites green.
- `duplicate-download.test.js`'s source assertion upgraded from `stampAids` to `prepareDocVersion`.

### DIFF
--- a/bin/tdoc-bake
+++ b/bin/tdoc-bake
@@ -43,7 +43,10 @@ function templateStamp(css) {
 }
 
 function hasBlock(html) {
-  return html.includes('id="tdoc-reader"');
+  // Tag match, not substring: a document whose prose quotes id="tdoc-reader"
+  // in a code sample must still be baked. Mirrors hasReaderBlock() in
+  // worker/worker.js.
+  return /<style[^>]*\bid="tdoc-reader"/i.test(html);
 }
 
 // Returns true when the file was changed.
--- a/bin/tdoc-publish
+++ b/bin/tdoc-publish
@@ -1198,8 +1198,13 @@ upload_version() {
   payload_file="$(mktemp -t tdoc-upload.XXXXXX)"
   printf '%s' "$payload" > "$payload_file"
   local resp
+  # Tier-1 client visibility: the server records which client produced each
+  # version. This is the observability that does not depend on the client's
+  # own (repeatedly silently-broken) self-update machinery.
+  TDOC_CLIENT_VERSION="$(cat "$SKILL_DIR/VERSION" 2>/dev/null || echo unknown)"
   resp="$(curl -sS --connect-timeout 10 --max-time 120 -X POST "$UPLOAD_BASE/api/upload" \
     -H "Authorization: Bearer $TOKEN" \
+    -H "X-Tdoc-Client: $TDOC_CLIENT_VERSION" \
     -H "Content-Type: application/json" \
     --data-binary @"$payload_file")" || { rm -f "$payload_file"; echo "[tdoc] v$v upload: network error/timeout" >&2; return 1; }
   rm -f "$payload_file"
--- a/server/server.js
+++ b/server/server.js
@@ -789,7 +789,10 @@ const server = http.createServer(async (req, res) => {
       // Roman. The proxy is unnecessary: the template is :where()
       // zero-specificity throughout, so a document that does style itself
       // wins every property it declares and is unaffected by the injection.
-      if (!body.includes('id="tdoc-reader"')) {
+      // Tag match, not substring: a doc whose PROSE quotes id="tdoc-reader"
+      // (tdoc's own design docs do) must still receive the template. Mirrors
+      // hasReaderBlock() in worker.js.
+      if (!/<style[^>]*\bid="tdoc-reader"/i.test(body)) {
         const rcss = readerCss();
         if (rcss) {
           const rtag = `<style id="tdoc-reader">${rcss}</style>`;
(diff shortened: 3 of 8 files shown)


serenakeyitan (approved):
Reviewed the full diff and ran the offline suite twice on the PR head (51/51 green; one flaky `cli.test.js` failure under the runner on the first pass passed 76/76 standalone and on re-run — not attributable to this change).

`prepareDocVersion()` is correctly wired into all three writers (/api/upload, browser-Save DO, duplicate), the widget upload loop stores raw bytes untouched, and the bake detection exactly mirrors the read-side `injectReaderCss` check, so no double-bake and no new inconsistency. `X-Tdoc-Client` is length-capped server-side. LGTM.

yayashuxue:
Both findings addressed in `87cfa52`.

**Copilot's substring false-positive — real, and closer to home than the example suggests:** tdoc's own design docs quote `id="tdoc-reader"` in code samples ([this one](https://tdoc.dev/d/tdoc-styling-map/v/7) does, several times), so an unbaked document *about the mechanism* would have been skipped by the write path. Every bake/skip decision now goes through one tag-matching test — `hasReaderBlock()`, `/<style[^>]*\bid="tdoc-reader"/i` — mirrored verbatim in `server/server.js`'s frame fallback and `bin/tdoc-bake`, so the write side and the read side cannot disagree. Test: a doc quoting the id in prose still gets exactly one real block.

**Serena's backfill-sha gap (raised on #342, fixed here where the recording lives):** non-latest uploads now refresh just that entry's `sha`/`client`, touching nothing else in meta; a failure there logs and never fails the upload — the bytes are stored and correct, a stale recorded sha costs at worst one spurious `/raw` re-download. Test corrupts the recorded sha, backfills, and asserts it matches the stored bytes again — and that the other version entries survive.

`write-invariants.test.js` is now 8 tests on this branch (12 with #342's four on top). Offline suite green.

yayashuxue:
## Design decision, for the record: the render fallback never writes storage (no read-repair)

A legacy document (stored before this PR) gets the reading template injected into **every** `/frame` response, and its stored bytes stay unbaked until its next authenticated write. That repetition is deliberate. We considered read-repair — bake into R2 on first read — and rejected it, four reasons in order of weight:

1. **Reads must have zero side effects.** `GET /frame` is anonymous and unauthenticated. If it can mutate storage, any visitor triggers writes. The system's whole posture is *only authenticated write paths touch storage*; read-repair cuts straight through that line.

2. **It races the real writers.** The write side is deliberately serialized: browser Save goes through the DO's version reservation + CAS, upload through the same reservation. A write emerging from the read path bypasses all of it and races a concurrent real save — last writer wins. And since this PR, *stored bytes ↔ recorded `sha`* is an invariant: read-repair would have to update meta's `sha` too, turning one GET into two racing writes.

3. **Blast radius.** The read path is the highest-traffic path. A bug in the fallback as it stands corrupts one response; a bug in read-repair corrupts storage, in bulk, from the busiest entry point.

4. **The economics don't justify it.** The repeated cost is a microsecond string check + splice; the saving would be one R2 write. And the population needing rescue shrinks on its own — every edit/publish goes through `prepareDocVersion()`, so storage self-heals one document per write.

The resulting asymmetry is the contract:

```
making storage better  →  only through an authenticated write
                          (edit, publish, or an explicit one-time backfill migration)
reads                  →  read-only forever; the fallback decorates the response, never the record
```

If we ever want to clear the legacy population actively, the correct shape is an explicit, authenticated backfill (a `tdoc-bake --remote`-style migration run once as a deliberate act) — not a side effect of reads. We chose **not** to run one for now: the consequences of the unbaked leftovers are benign (a few extra microseconds per render), and next-write self-healing retires it naturally.

PR #342, opened 2026-09-01
An agent reads the source of truth, one conditional request at a time

yayashuxue:
> Stacked on #341 (needs its `sha`). Read side of the same design: [文档的样式从哪来 §11](https://tdoc.dev/d/tdoc-styling-map/v/7).

## The gap

`AGENTS.md` line one: remote storage is the source of truth. The `/tdoc edit` procedure: "read the latest version's `index.html`" — the local disposable copy. `tdoc-pull` already conceded this mismatch for comments; HTML had no equivalent. `tdoc-next` showed the cost: remote v2 carried the baked template, local v2 was 26 bytes, and the editing agent faithfully imitated the stale file.

## The handshake

```
publish   →  upload response returns sha of the stored bytes
          →  tdoc-publish records it on the local meta.json version entry
edit      →  HEAD /d/<slug>/v/<n>/raw  →  ETag
             match     → local copy produced what remote holds; use it (zero bytes)
             differ    → GET /raw, base the edit on the truth
             unreachable → use local, and say so in the reply
```

The recorded sha, not a recomputed one, is the comparison key — the local file hashes differently by design (it is the pre-stamp, pre-bake form).

## `/raw` security shape

Author HTML on the shared origin outside the sandboxed `/frame` would be stored XSS, so:
- `Content-Type: text/plain` + `nosniff`, never `text/html`
- same `enforceDocAccess` gate as every other doc read (test: anonymous read of a private doc denied, owner token passes)
- `/raw` never bakes — it reports the truth, it does not improve it (test pins legacy bytes returned untouched)

## Also

`X-Tdoc-Client` now carries the skill checkout's **git sha** — the `VERSION` file has said `0.9.0` since the beginning and cannot distinguish any two clients in the wild. This is the observability that attributes a broken published doc to the exact skill code that produced it.

## Verification

Four new tests in `write-invariants.test.js` (raw bytes/ETag/content-type; 304 with no body; computed ETag for legacy storage; access gating). 51/51 offline suites green; `cli.test.js` 76/76.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -532,7 +532,24 @@ comments you handled unless you reply on each one. Skipping comments
 silently is the #1 source of regression complaints.
 
 1. Read `~/tdocs/<slug>/comments.json` — filter to `status: "open"`.
-2. Read latest version's `index.html`, and re-read `$SKILL_DIR/authoring/voice.md`, `$SKILL_DIR/authoring/visuals.md` and `$SKILL_DIR/authoring/structure/components.md`.
+2. **Get the current document — remote is the source of truth, local is a
+   cache.** The local `v<n>/index.html` can be stale: a browser edit or a
+   publish from another machine creates versions your checkout never saw, and
+   an edit based on a stale copy silently discards them. One conditional
+   request settles it (`published.json` holds the base URL; skip this entirely
+   for a doc that was never published):
+
+   ```bash
+   REMOTE_SHA="$(curl -sfI "$BASE/d/<slug>/v/<n>/raw" | tr -d '\r' | sed -n 's/^etag: "\(.*\)"$/\1/Ip')"
+   LOCAL_SHA="$(node -e 'const m=require(process.argv[1]);const e=(m.versions||[]).find(v=>v.n===Number(process.argv[2]));console.log(e&&e.sha||"")' "$TDOC_DIR/<slug>/meta.json" <n>)"
+   ```
+
+   - **Match** → your local copy produced what remote holds; use it as the base.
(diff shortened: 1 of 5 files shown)


serenakeyitan (approved):
Reviewed the full diff and ran the offline suite on the PR head (51/51 green). The `/raw` route is gated with the same `enforceDocAccess` as every other doc read, served `text/plain` + `nosniff`, and the ETag/304 handshake is correct for GET and HEAD, with a computed-hash fallback for legacy storage. Verified the SKILL.md sed one-liner works on macOS. Git-sha client reporting is the right call over the frozen VERSION file. LGTM.

One minor gap, non-blocking: `entry.sha` is only recorded when the upload writes latest meta (the `writesLatestMeta` gate in /api/upload). A history backfill (`--history` re-uploading v1..vN-1) stores freshly-baked bytes but never refreshes those entries' shas — so after a reader-template generation change, `/raw` for an old version can serve a stale recorded ETag that doesn't match the actual stored bytes. Consequences are benign (worst case: an edit bases on a content-equivalent pre-bake copy, or one spurious re-download), but it slightly weakens the "recorded sha = truth" invariant. A one-liner recording sha on non-latest-meta writes too — or a comment documenting the gap — would close it.

yayashuxue:
The backfill-sha gap you flagged is closed in #341's `87cfa52` (the recording lives there): non-latest uploads refresh the entry's `sha`/`client` in place, nothing else in meta is touched, and a failure logs without failing the upload. Regression test corrupts a recorded sha, re-uploads the old version, and asserts the entry matches the stored bytes again.

This branch is rebased on top and re-verified — `write-invariants.test.js` 12/12, including the four `/raw` tests. One side effect worth noting: with `hasReaderBlock()` from #341, `/raw`'s "never bakes" property now also holds for documents whose prose quotes the id — same tag-match everywhere.

Issue #349, opened 2026-09-01
Comments: reply box stays open after submit, plus edit / delete permissions / AI re-commenting


yayashuxue:
A batch of comment-system problems from user feedback (2026-09-01). The main one is the reply box never resetting after a successful submit; the rest are permission/interaction gaps.

## 1. Reply box stays open after submitting (bug)

Repro: click **Reply** on a comment → type a reply → submit.

Actual: the reply posts, but the composer stays on screen with its text still in it. Only a page refresh clears it.

Expected: on a successful submit, close the composer, clear the draft, and insert the new reply into the thread in place — no refresh needed.

## 2. Let a comment's author edit their own comment (enhancement)

Comments are immutable once posted. The author of a comment should be able to edit it (nobody else). An edited comment should carry an "edited" marker.

## 3. The doc author can delete other people's comments (needs a product call)

Today the doc author can delete comments left by other people, which feels wrong: deletion should belong to whoever wrote the comment. The right affordance for a doc author is resolve/hide, not deleting someone else's words. Decide which model we want before changing code.

## 4. AI re-comments on spots a human already handled (bug / UX)

After a human deletes or edits an AI comment, the next generation round drops the same comment back in the same place.

Expected: the AI should only comment where it hasn't commented before; anchors a human has already dealt with (deleted, edited, resolved) should not be re-raised.

---

Source: user feedback over chat, translated from the original Chinese:

> "The box is still there after I type and submit / it's gone after a refresh — must be a bug"
> "Also add edit comment; the author should be able to edit"
> "And I'm not a fan of the doc author being able to delete other people's comments — that feels off"
> "The AI should only comment where it hasn't commented already"

serenakeyitan:
Split into four, one per problem, all open and **unmerged** pending Serena's manual pass. Each PR body ends with a click-by-click local recipe.

| # | PR | Base |
|---|----|------|
| 1. Reply box stays open | #350 — A posted reply closes its box and shows up in the thread | `main` |
| 2. Author can edit their own | #351 — The author of a comment can rewrite it, and it says so | `main` |
| 3. Delete permissions | #352 — Deleting a comment belongs to whoever wrote it | **stacked on #351** |
| 4. AI re-comments | #353 — The agent answers a comment once per human turn | `main` |

Two notes:

[...]

PR #352, opened 2026-09-01
Deleting a comment belongs to whoever wrote it

serenakeyitan:
Third of four splits of #349 — the "doc author can delete other people's comments" product call.

> **Your call, verbatim:** *"deletion should belong to whoever wrote the comment."* That is what this implements. It does **not** add the resolve/hide affordance the issue floats as the owner's replacement — see *What this costs* below, and say the word if you want it.

**Stacked on #351** (`feat/edit-your-own-comment`), because delete and edit end up behind the same "is this your own" gate and splitting them across two bases would mean writing that gate twice. Merge #351 first; this rebases onto `main` cleanly after.

## The change

`canMutate()` — author **or** doc owner — gated three things. It keeps the one it is right for: **re-anchoring** is about where a comment points, not what it says, and an owner rebinding a stale anchor after a republish is repair work, not censorship. **Delete** leaves it and joins edit behind the author's own gate — `isRecordAuthor()` in the worker, `authoredBy()` in the card, both denying on a null-author legacy record rather than short-circuiting to allow.

The DELETE route also gets shorter: it used to resolve the target twice (once as a comment, once by scanning every thread's `reply_added` events) with an `authorized` flag threaded through both branches. `findRecord()` from #351 already does that resolution, so the route is now resolve → check → write.

`ReplyCard` stops taking `isOwner`: a reply row now has nothing the doc owner unlocks.

## An agent's words belong to the doc it was run on

Author-only delete had a hole that only showed up next to #353: an agent reply's author is the *agent*, so under the new rule **nobody** could delete one. Merging the two branches proved it — the owner deleting Claude's reply came back `403 not_author`, and the case #349 point 4 is entirely about ("after a human deletes an AI comment") stopped being possible.

That is not an exception to the rule, it is the rule asked one level down: **whose words are an agent's?** `/api/agent/reply` is authed with the doc's **upload token** — the owner's publishing credential. The agent is not a third party with speech of its own; it is the owner writing through a tool.

So `mayDelete()` reads: **your own words, plus an agent's if the doc is yours.** `isRecordAuthor` stays exactly as strict for everything a person wrote, and **edit is untouched** — nobody rewrites what the agent said under its name, the owner included. On a hosted doc it follows the publisher rather than `TDOC_OWNER`.

Settled with @serenakeyitan; `#353` + this branch together are green (13/13).

## What this costs (flagging, not arguing)

- An owner keeps only the blunt instruments against a *human's* comment: delete the document, or `DELETE ?all=1` (the upload-token wipe the publish tooling holds).
- **Nobody** can delete a comment whose author is gone — legacy records with no author, or an account that no longer resolves. Previously the owner could clean those up. (A null-author record is deliberately *not* an agent record, so the carve-out does not reach it.)

If either bites, the fix is #357 — resolve — not widening delete.

## Testing

The security suite now states the new rule where it stated the old one:
- the card renders delete only behind `authoredBy`, on both the comment and every reply, and behind nothing that mentions the owner
- the worker's DELETE branch checks `isRecordAuthor` and never calls `canMutate`
- `mayMutate` is declared once and called once — for the re-anchor

The browser test that proved the owner still had a delete on someone else's comment now proves the opposite, and that the re-anchor is what remains.

`npm test`: 53/53 green. `artifact-shell` (browser, gated): 31 passed with the same 4 failures `main` already has.

**Do not merge — pending manual testing by @serenakeyitan.**

Refs #349

---

## Test it by hand

> The PR preview Worker is failing on **every** open PR right now (`version_conflict` seeding the landing doc — a `main` regression, not this branch), so this is the local recipe.

```bash
git fetch origin && git checkout fix/delete-belongs-to-the-author
npm install && npm run build:shell
TDOC_E2E_USER=tester TDOC_DIR="$PWD/test/fixtures/tdocs" TDOC_PORT=7899 node server/server.js
```

Then open **http://localhost:7899/d/hostile-body-css/v/1**.

The fixture doc is the surface to use because its comments belong to four different people — `tester`, `tester2`, `reviewer-a`, `reviewer-b`, and an agent — so the permission split is visible without a second GitHub account. `TDOC_E2E_USER=tester` signs you in as `tester`, who wrote `c_fixture_1` and also counts as the doc owner locally.

The fixture's `comments.json` is a real file the server writes to. `git diff test/fixtures/` when you are done, and revert it if you want the checkout clean.

### What to check

1. Open `c_fixture_1` (yours as `tester`) — `edit` and `delete` are both there. Delete still works on your own.
2. Open *"first reviewer: this sentence is confusing"* (`reviewer-a`'s). You are the **doc owner** and there is now **no delete** on it — and no edit. What remains is `↻ move anchor`, which is deliberate: re-anchoring is about where a comment points, not what it says.
2b. Open *"fixed this in the latest revision"* (Claude's). You **do** get a delete on that one — an agent writes with your upload token — and still no edit: nobody rewrites what it said.
3. Compare against `main` on the same card: `delete` is there today.
4. Expand `c_fixture_1`'s thread — `tester2`'s reply has no delete for you either.

Worth deciding while you have it open: an owner is now left with only the blunt instruments (delete the doc, or the `?all=1` token wipe), and **nobody** can delete a comment whose author is gone. If that bites, the answer is the resolve/hide affordance the issue floats — say the word and it becomes its own issue.


serenakeyitan:
Closed by GitHub, not on purpose: deleting the base branch when #351 merged auto-closed this, and a closed PR cannot be retargeted or reopened once its base is gone. The work continues at #359 — the same two commits cherry-picked onto `main`, content unchanged.

PR #386, opened 2026-09-02
An invite adds a name and changes nothing else

### LINKED ISSUE #385: An @mention invite silently makes a public doc unlisted and hides its history
Introduced by #348, merged 2026-09-01. Posting a comment can silently rewrite a document's sharing policy.

A doc published without an `access` object — every CLI-published doc — reads as public with public version history. `applyAccessPatch` deliberately switches such a doc to the product defaults on its first policy write (`worker.js:246`), which is correct when the owner is editing the Share panel and is the only thing that used to call it.

#348 added `@`-mention-as-invite, which calls the same function to widen `allowed_users` — as a side effect of posting a comment:

```
before:  {"visibility":"public",   "history_visibility":"public"}
after:   {"visibility":"unlisted", "history_visibility":"owner"}
```

So an owner who `@`-mentions someone not already on the doc silently takes the doc out of public and hides every earlier version from anyone holding the share link. Reported as "why can't people see previous versions after I share".

Blast radius: any doc with no stored `access` object that an owner `@`-mentioned a non-participant on since #348 merged.

Fix: the invite passes the doc's effective policy through, so the re-baseline is a no-op and an invite adds a name and changes nothing else. The Share-panel path keeps today's behavior.

serenakeyitan:
Fixes #385. Regression from #348, merged yesterday.

A doc published without an `access` object — every CLI-published doc — reads as public with public version history. `applyAccessPatch` deliberately re-baselines such a doc to the product defaults on its first policy write. That is right when the owner is editing the Share panel, and it used to be the only caller.

#348's @mention-as-invite calls the same function to widen `allowed_users`, as a side effect of posting a comment:

```
before:  {"visibility":"public",   "history_visibility":"public"}
after:   {"visibility":"unlisted", "history_visibility":"owner"}
```

An owner who @mentioned someone not already on the doc silently took the doc out of public and hid every earlier version from anyone holding the share link.

## The change

The invite passes the doc's effective policy through, so the re-baseline is a no-op — an invite adds a name and changes nothing else. Share panel behavior is untouched.

One call site in `worker/worker.js`. No new defaults, no new options, no behavior added.

## Tests

- the invite leaves visibility and history alone on a doc with no stored access
- the bare `allowed_users` patch still re-baselines — guards the diagnosis, so this call site can be simplified if `applyAccessPatch` ever changes

`npm test` — **58/58 offline suites green**, mentions 33/33.

## Not included

Docs already rewritten by the bug keep their stamped policy; this only stops new ones. Repairing those is a data question, not a code one.

### DIFF
--- a/test/mentions.test.js
+++ b/test/mentions.test.js
@@ -60,6 +60,14 @@ vm.runInContext([
   fn(src, 'positionalRecipient'),
   fn(src, 'inboxRecipients'),
   fn(src, 'inboxGroupKey'),
+  constLine(src, 'ACCESS_VISIBILITIES'),
+  constLine(src, 'ACCESS_COMMENTING'),
+  constLine(src, 'ACCESS_HISTORY'),
+  constLine(src, 'ACCESS_PATCH_FIELDS'),
+  fn(src, 'normalizeAccess'),
+  fn(src, 'accessFromMeta'),
+  fn(src, 'validateAccessWrite'),
+  fn(src, 'applyAccessPatch'),
 ].join('\n\n'), box);
 
 // The shell's pure half is an ES module; strip the export keyword and run the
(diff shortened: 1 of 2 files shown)


serenakeyitan:
Closing: the premise is wrong, and I verified it against a real worker.

The invite branch only runs for logins that **cannot read** the doc. A doc with no stored `access` reads as public, so everyone can read it, so `invited` is always empty and `applyAccessPatch` is never called. The branch can only fire on a **private** doc — and a doc is only private because someone explicitly set a policy, which materializes all four fields, so the re-baseline is already a no-op there.

Checked end to end on the pre-fix bundle running under `wrangler dev`: owner @-mentions a stranger on a 3-version legacy doc, and afterwards `meta.access` is still `null`. No write happened, and an anonymous reader still gets all 3 versions.

My earlier evidence was a hand-built call to `applyAccessPatch` that no route can produce. Sorry for the noise.

PR #399, opened 2026-09-02
The identity model, done properly: email people participate, recycled identifiers do not inherit

yayashuxue:
Two related pieces of the identity model, in one PR because they touch the same functions and the second one is a **live security fix**.

Design doc: https://tdoc.dev/d/tdoc-auth-refactor/v/5

---

## Part 1 — an email-only person is a full participant

Phases 1–3 let an OIDC visitor sign in and publish, then stopped: the gates deciding who may **comment**, who an **invite** matches, and who **owns their own words** all asked for a GitHub handle. So an email identity could publish a doc and not leave a single comment on it — the one thing tdoc exists for.

The mechanism is an **actor key**: a GitHub identity keeps its bare handle (every comment, inbox entry and allowlist ever written uses that shape), an identity with no handle takes an `email:` prefix. The namespaces cannot collide in either direction. Everything keyed on "who" uses it — comment authorship, both authorship gates, event actor stamps, inbox keys, stars/recents/folders.

Two bugs found on the way:

- **`allowed_users` write validation still rejected addresses**, so D2's email invites could not be stored at all. Detection is handle-first, not "contains @" — `@Bob` is a handle wearing a sigil, and treating it as an address broke three existing cases.
- **The private-doc guard used `sessionLogin`**, so a signed-in email visitor was told to *sign in* (401) rather than that they are not on the list.

Addresses never render where readers can see them — an author's display name falls back to the local part, and the access-denied page no longer echoes the visitor back at them.

---

## Part 2 — recycled identifiers must not inherit an account

Two things a person controls can change hands, and both were load-bearing keys:

**A GitHub login can be renamed, and the freed name registered by anyone.** Accounts were keyed on `hosted-account:<login>`:

1. Alice publishes as `@alice`
2. Alice renames to `@alice-dev`
3. Bob registers the freed `@alice`
4. **Bob signs in and gets Alice's account and documents**

No vendor involved — this was live. `ghUser` had been returning GitHub's immutable numeric id all along and nothing ever read it.

**An email address gets reassigned.** A company hands a departed employee's mailbox to someone new, who verifies it at any provider. The email index was written and never retracted, so it kept pointing at the old holder's account.

### The fix

```
account-idp:<provider>:<sub>  → account_id   authoritative — immutable, never reissued
hosted-account:<login>        → account_id   legacy; only claimable if it has no stable owner
account-email:<email>         → account_id   a merge hint: used once, then superseded
```

Stale pointers are retracted when the identity that wrote them moves on. A handle index whose account already has a stable GitHub owner is no longer claimable by a different id.

### This reverses a rule asserted earlier in this same branch

An earlier commit said the issuer's `sub` must never be stored, because a vendor's id in our data is lock-in. **That had it backwards.** Lock-in is about who owns the *account*: `account_id` is ours and every doc, token and permission hangs off it. Drop a provider and its idp index becomes dead weight — the account is untouched, and its owner walks back in through the email hint on whatever replaces it. Refusing to store `sub` bought nothing and cost the one guarantee an address cannot give.

Accounts now also record their identities (`provider`, `sub`, `handle`, `email`, `linked_at`, `last_seen`) — what an account page will read, and what answers "how did this person get in" when something looks wrong.

**Migration is lazy and total**: a legacy account resolves by handle exactly as before, and that same sign-in writes its stable index. Nobody is moved.

---

## Verification

`test/email-identity-surface.test.js` (9) — namespaces cannot collide; email identities comment, edit only their own, are invited without borrowing handle invites; addresses never render; stars land under `stars:email:…` instead of vanishing.

`test/identity-recycling.test.js` (6) — drives real sign-ins against stubbed providers:

- a renamed login keeps the account; **the freed name gives Bob nothing**
- a recycled address **does not inherit** the account it used to reach
- changing your address at the provider **keeps** your account
- the email hint still merges a new provider into an existing account, once
- an account records how its owner signs in
- **legacy accounts upgrade on next sign-in**

Full offline suite: **69/69 green**.

https://claude.ai/code/session_013Eow4sjmakK5n7JkFVhqDh


yayashuxue:
Superseded by #408, which carries these commits plus the identity hardening. This branch never got a `pull_request` dispatch — the cause turned out to be that it was behind main, so GitHub could not compute mergeability and never fired the event.

PR #419, opened 2026-09-03
The comment's action row stops being five coloured words

### LINKED ISSUE #416: The comment card's action row is five coloured words in a line
The row of controls under a comment has grown one action at a time and now
reads as a strip of coloured words:

```
☺  reply  edit  reopen  delete
```

Five controls, three colours — blue for reply and edit, green for reopen, red
for delete — all bare 11px text with no shared shape, sitting next to a
graphical icon that is the only thing in the row with a hit area. Nothing
groups them, nothing separates the ordinary actions from the destructive one,
and the more the card can do the worse it gets: `resolve` was the fifth, and
there is no obvious place to put a sixth.

The colours are doing work they should not. Red on `delete` is meaningful; blue
on `reply` and `edit` is not — they are just links because that is what the
legacy overlay made them, and `reopen` picked up green only because it matched
the resolved chip.

## What to look at

- `.tdoc-margin-comment .meta` / `.actions` in `server/chrome.css` — the
  layout
- the button-reset block in `shell/src/ui/ui.css` (`button.del`,
  `button.tdoc-reply-toggle`, `button.tdoc-edit-toggle`,
  `button.tdoc-resolve-toggle`) — the per-control colours
- `CommentCard` in `shell/src/document/comment-card.jsx` — the reply row has
  the same problem one level down

## Constraints worth keeping

- The row has to stay quiet: a comment card is mostly text, and the controls
  should not compete with it until you look at them.
- The reaction picker already appears on hover only when there are no
  reactions. Whatever this becomes should keep that restraint.
- Both levels — a comment and a reply — carry the same set, so one treatment
  has to serve both.
- Cards are narrow (280px) and the set is longer on a thread you own, so
  whatever it becomes has to survive wrapping.

## Not in scope

Which controls exist, or who may press them. This is about how the row reads,
not about changing what it can do.

serenakeyitan:
Fixes #416.

```
before   ☺  reply  edit  reopen  delete      blue · blue · green · red
after    ☺  reply  edit  resolve   delete    grey · grey · grey  ·  muted red
```

`reply` and `edit` were blue because the legacy overlay made them links.
`reopen` was green because it matched the resolved chip. `delete` was red. Five
controls, three colours — and only the red was ever saying anything.

## What changed

**One resting voice.** Everything you can do to a thread is the same grey and
darkens on hover *or keyboard focus*. A card is mostly text; the row should
stay out of the way until it is looked at.

**Delete keeps a colour**, because it is the one action that does not come
back. It sits at a muted `#c9776f` and deepens to `#c33` on hover, rather than
turning into a different word when you approach it.

**A gap does the work a colour would.** 6px pushes delete off the end of the
run, so it is not the fifth word in a list of five equal ones. The row wraps,
because a 280px card with five controls has to be allowed to.

**The words got a target bigger than the words.** 11px text with no padding is
a small thing to hit. Padding grows the box and a matching negative margin
gives the space back — same appearance, `13px → 23px` tall.

## Why it is one rule and not five

The rule is scoped to the card (`.tdoc-margin-comment .actions button`) rather
than to each control, so a **reply's** row inherits it with no second copy to
keep in step. That is also what a test asserts: a reply-specific copy of this
rule is exactly how the two levels would drift apart.

## Verification

`npm test` — 73/73 suites green, including the new `comment-actions-row.test.js`.

Measured live, not eyeballed — computed colour and box per control:

```
comment:  icon 26px #8a8a8f · reply 23px #8a8a8f · edit 23px #8a8a8f
          · resolve 23px #8a8a8f · delete 23px #c9776f (margin-left 6px)
reply:    icon 26px #8a8a8f · reply 23px #8a8a8f · edit 23px #8a8a8f
          · delete 23px #c9776f
```

## Not in scope

Which controls exist, or who may press them — #416 was about how the row reads.

---
<sub>🌱 PR created by Claude on behalf of @serenakeyitan via <a href="https://claude.com/claude-code">Claude Code</a></sub>


serenakeyitan:
Closing unmerged — the owner looked at it and it reads worse than what it replaced, so the change is not worth keeping.

Flattening reply/edit/resolve to one grey removed the colour but did not give the row any structure back, and the muted red plus the 6px gap made delete look detached rather than deliberate. The hit-area padding was the only part that stood on its own.

Leaving #416 as filed: the observation still holds, this particular answer to it does not.

PR #423, opened 2026-09-03
A table wider than its container is not a phone problem

### LINKED ISSUE #422: The table rule stops at 700px, so tablets and narrowed windows still pan
#417 scoped the table rule to `@media (max-width: 700px)` — the reasoning was "don't touch desktop". That boundary is wrong: what breaks a page is a table wider than its container, and a container gets narrower than a table on a tablet and in a narrowed desktop window too.

Swept a real doc across widths against the deployed fix:

| viewport | document overflow |
|---|---|
| 320–700px | 0 ✓ |
| **701px** | **171px ✗** |
| **768px** (tablet, phone landscape) | **164px ✗** |
| **834px** | **115px ✗** |
| **1024px** | **20px ✗** |

The author's `table{min-width:880px}` goes unchecked above 700, so everything from a tablet up still pans sideways.

## Fix

Drop the media query. The rule is self-limiting: a table that already fits its container is laid out exactly as before, because `max-width:100%` and `min-width:100%` on the rows only bite when the content is wider. Nothing about it is phone-specific.

Swept again on the change — four docs × eleven widths from 320 to 1280, all zero:

```
rl-data-landscape          320:✓ 375:✓ 390:✓ 430:✓ 600:✓ 700:✓ 701:✓ 768:✓ 834:✓ 1024:✓ 1280:✓
tdoc-sys-paper             (same)
santa-cruz-surf-roadtrip   (same)
overflow-repro             (same)
```

Column widths stay sensible and grow with the container: `[64,224,220]` at 375px, `[92,292,288]` at 1024px.

The drift test now asserts the **absence** of a media query, with the measured numbers in the comment, so the scoping cannot quietly come back.

serenakeyitan:
Fixes #422. Follow-up to #418, found by sweeping every responsive width instead of only the one that was reported.

#418 scoped the table rule to `@media (max-width: 700px)` — "don't touch desktop". That boundary is wrong: what breaks a page is **a table wider than its container**, and a container gets narrower than a table on a tablet and in a narrowed desktop window too.

## Measured against the deployed fix

| viewport | document overflow |
|---|---|
| 320–700px | 0 ✓ |
| **701px** | **171px** ✗ |
| **768px** — tablet, phone landscape | **164px** ✗ |
| **834px** | **115px** ✗ |
| **1024px** | **20px** ✗ |

The author's `table{min-width:880px}` went unchecked above 700.

## The change

Drop the media query. The rule is self-limiting — a table that already fits is laid out exactly as before, because `max-width:100%` on the block and `min-width:100%` on the rows only bite when the content is wider. Nothing about it was ever phone-specific.

Swept again on this branch — four docs × eleven widths, 320 → 1280:

```
rl-data-landscape          320:✓ 375:✓ 390:✓ 430:✓ 600:✓ 700:✓ 701:✓ 768:✓ 834:✓ 1024:✓ 1280:✓
tdoc-sys-paper             (same)
santa-cruz-surf-roadtrip   (same)
overflow-repro             (same)
```

Column widths stay sensible and grow with the container: `[64,224,220]` at 375px, `[92,292,288]` at 1024px.

The drift test now asserts the **absence** of a media query, with the measured numbers in the comment, so the scoping cannot quietly come back.

`npm test` — **73/73 offline green**; browser-editing 17/17, responsive 32/32.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -641,7 +641,7 @@ function frameCspHeader(nonce) {
 // documents bake their reader CSS at creation: changing that file alone fixes
 // nothing that is already published. Kept byte-identical in server.js —
 // test/reader-patch-drift.test.js holds the two together.
-const READER_PATCH_CSS = '@media (max-width:700px){body table:not(.tdoc-table-scroll>table){display:block!important;min-width:0!important;max-width:100%!important;overflow-x:auto;-webkit-overflow-scrolling:touch}body table:not(.tdoc-table-scroll>table)>thead,body table:not(.tdoc-table-scroll>table)>tbody,body table:not(.tdoc-table-scroll>table)>tfoot{display:table!important;width:max-content!important;min-width:100%!important}}';
+const READER_PATCH_CSS = 'body table:not(.tdoc-table-scroll>table){display:block!important;min-width:0!important;max-width:100%!important;overflow-x:auto;-webkit-overflow-scrolling:touch}body table:not(.tdoc-table-scroll>table)>thead,body table:not(.tdoc-table-scroll>table)>tbody,body table:not(.tdoc-table-scroll>table)>tfoot{display:table!important;width:max-content!important;min-width:100%!important}';
 
 const READER_CSS_PATH = path.join(__dirname, 'reader.css');
 function readerCss() {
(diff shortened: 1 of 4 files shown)


server/server.js
mccrearydamion69-crypto: Hi

worker/worker.js
mccrearydamion69-crypto: Good morning

mccrearydamion69-crypto: 1damionmccreary@cloud.com

mccrearydamion69-crypto: What up

mccrearydamion69-crypto: Hi

PR #552, opened 2026-09-14
The desk is a desk, the slug is theirs, and the button is on the screen

serenakeyitan:
Three things the right-hand replay and the column beside it were getting wrong, all visible on the first screen a new reader sees.

## The desk

- **The app was open before it was opened.** Lifting the window into a shared component dropped its entrance, so it was drawn from frame 0 — the pointer was still walking to a dock whose app was already up.
- **The desktop was a purple wash** no screen has. Now the blues macOS actually ships.
- **The menu bar was a dark strip with one word on it.** Now the apple, the front app in bold, its menus, and the status items.
- **The dock held three chat apps and nothing else** — which is not a dock anybody has. Now Finder / Safari / Messages, then the three this story is about, then the bin behind a separator.
- Before there is a line to paste, the desk is there and no app is open — true, and the *same* desk, rather than a second window in a second style.

## The slug

The replay published `what-standups-cost` no matter what the reader typed, so someone asking for a doc about a trip watched a doc about standups get published. It travels with the line now — the two are one take.

## The button

Measured on tdoc.dev at 863px: **753px needed**, of which 126 was seven 18px gaps and 64 was the strip the internal bar reserves. Below ~760px the only action on the page goes under the fold.

The collapse that was supposed to prevent this already existed — and never once ran:

```css
@media (max-height: 870px) { .sg-col { gap: 15px; } }   /* written first */
.sg-col { …; gap: 18px; }                                /* base rule, written after */
```

Same specificity, later in the file, so the base won at every height. It measured 18px on the very window it was written for. Base first, overrides after, and the air now comes out in three steps (870 / 790 / 710) while the rows stay the size they are. The internal bar's reserved strip yields too, rather than pushing the product's own button off the screen.


serenakeyitan:
Closing and reopening as a fresh PR: this one never triggered a single workflow run from the day it was created (0 runs on the branch via the Actions API, while the PR itself showed a normal OPEN state pointing at the right commit). Reopening it and pushing an empty commit both failed to wake CI. Same branch, same commits — see the replacement.

